Issue #9, opened 2021-06-11
List.repeat simplification is wrong


gampleman:
```
-- ELM-REVIEW ERROR ---------------------- src/elm-extras/List/Extras.elm:173:34

Simplify: Using List.repeat with an empty list will result in a empty list

172| transpose matrix =
173|     List.foldr (List.map2 (::)) (List.repeat (rowsLength listOfLists) []) listOfLists
                                      ^^^^^^^^^^^

You can replace this call by an empty list.

I think I can fix this. Here is my proposal:

172| transpose matrix =
173|     List.foldr (List.map2 (::)) (List.repeat (rowsLength listOfLists) []) listOfLists
174|     List.foldr (List.map2 (::)) ([]) listOfLists
174|

? Do you wish to apply this fix? › (Y/n)
```

This is wrong:

```
ampleman@MacBook-Pro ui % elm repl
---- Elm 0.19.1 ----------------------------------------------------------------
Say :help for help and :exit to exit! More at <https://elm-lang.org/0.19.1/repl>
--------------------------------------------------------------------------------
> List.repeat 3 []
[[],[],[]] : List (List a)
>
```

sparksp:
@jfmengels could/should this be extended for any value repeated a known number of times? Or would that introduce a possible readibility issue?

jfmengels:
I think things like `List.repeat 100 a` transformed to `[ a, a, a, ...]` would create a readability issue, yes. It can also become a performance issue when changing something like `List.repeat n (expensive a)` to `[ expensive a, expensive a, ... ]` (we will not likely create a let declaration)

The same question has been asked about `String.repeat 100 "a"`, which I don't think will make for nicer code. Also, it's easy to tell at a glance that the string contains 100 characters, which is much harder to do with a raw string, and which might be relevant to the code in question (like inside a test)

The rule should remove unnecessary operations, but not pre-compute the result when the aim of the user was likely to make the code nicer. So `String.repeat 1 "a"` will be simplified to `"a"`, but not for higher values than 1.

sparksp:
With `String.repeat` in mind, should the handling for `List.repeat` with `n` > 0 be removed entirely? `[ [], [], [], ... ]` is essentially pre-computation?

jfmengels:
We are not currently changing List.repeat by `[ [], [], [], ... ]`. I am confused by what you're asking :thinking:

sparksp:
[...]

PR #10, opened 2021-07-05
Fixes bad rule for List.repeat

gampleman:
As mentioned in #9, `List.repeat 3 []` is not equal to `[]` as the rule suggests, but rather equal to `[ [], [], [] ]`.


jfmengels:
Oh I totally missed this issue, my bad!

Thanks for the PR, it looks great!
I'll release this now :)

jfmengels:
And sorry for the inconvenience! :(

PR #20, opened 2022-03-31
Properly determine the size of sets

### LINKED ISSUE #19: Incorrect suggestion for `Set.size (Set.fromList [ x, y, z ])`
Example code:

```elm
triangleKind x y z =
    case Set.size (Set.fromList [ x, y, z ]) of
        1 ->
            Equilateral
         2 ->
            Isosceles
        _ ->
            Scalene
```

The suggestion is

```
Simplify: The size of the set is 3

20|     triangleKind x y z =
21|         case Set.size (Set.fromList [ x, y, z ]) of
                 ^^^^^^^^
22|             1 ->

The size of the set can be determined by looking at the code.
```

The suggestion is incorrect since `[x, y, z]` could be a smaller set than size 3.

The check could be dropped, or refined to trigger only for lists of literal values and the size would be obtained from`Set.fromList >> Set.size`.

I'm happy to do a PR, please advise :)

jiegillet:
Closes #19.

Full disclosure: this code is a little bit of a mess, so I will absolutely understand if you would prefer simply removing this check from the list.

The first commit is a real abomination, because I tried to use `Set.fromList >> Set.size` from a `List (Node Expression)`, but of course `Set` can only accommodate a comparable value, so I had to transform the expressions into a comparable type. It's bad, but possibly in a "it's so bad it's good" kind of way, which is why I left the commit.

The second commit defines a `unique` function (ripped from `List.Extra`) that is less efficient but less demanding. But then I had to do an extra step of reducing a `Node Expression` to an `Expression` to get rid of ranges. It's less horrible, but still messy.

Again, if you think it's not worth it, I can just skip the check for sets.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -4346,13 +4346,91 @@ determineIfCollectionIsEmpty moduleName singletonNumberOfArgs lookupTable node =
 
             Expression.Application ((Node fnRange (Expression.FunctionOrValue _ "fromList")) :: (Node _ (Expression.ListExpr list)) :: []) ->
                 if ModuleNameLookupTable.moduleNameAt lookupTable fnRange == Just moduleName then
-                    Just (Exactly (List.length list))
+                    case ( moduleName, list, List.all isComparableValue list ) of
+                        ( [ "Set" ], [], _ ) ->
+                            Just (Exactly 0)
+
+                        ( [ "Set" ], [ _ ], _ ) ->
+                            Just (Exactly 1)
+
+                        ( [ "Set" ], _, True ) ->
+                            list |> List.filterMap extractComparable |> Set.fromList |> Set.size |> Exactly |> Just
+
+                        ( [ "Set" ], _, False ) ->
+                            Just NotEmpty
+
+                        _ ->
+                            Just (Exactly (List.length list))
 
                 else
                     Nothing
 
             _ ->
                 Nothing
+
+
+isComparableValue : Node Expression -> Bool
+isComparableValue (Node _ expression) =
+    case expression of
+        Expression.Integer _ ->
+            True
+
+        Expression.Hex _ ->
+            True
+
+        Expression.Floatable _ ->
+            True
+
+        Expression.Negation expr ->
+            isComparableValue expr
+
+        Expression.Literal _ ->
+            True
+
+        Expression.CharLiteral _ ->
+            True
+
+        Expression.ParenthesizedExpression expr ->
+            isComparableValue expr
+
+        _ ->
+            False
+
+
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -4355,6 +4369,92 @@ determineIfCollectionIsEmpty moduleName singletonNumberOfArgs lookupTable node =
...
+            False
+
+
+simplifyComparableExpression : Node Expression -> List Expression
+simplifyComparableExpression =
+    simplifyComparableExpressionHelper 1
+
+
+simplifyComparableExpressionHelper : Int -> Node Expression -> List Expression
+simplifyComparableExpressionHelper sign (Node _ expression) =
```
jfmengels, line 4412: I think it would be interesting to also support `String`s. Maybe you can combine `simplifyComparableExpression/simplifyComparableExpressionHelper` and `isComparableValue` into one function that folds over expressions and depending on whether we've seen strings or numbers we combine them?

You could "parse don't validate" so that you don't iterate over the list twice. Also, you can use `Normalize.getNumberValue` from this repo to do the same thing you're doing here maybe?

tests/SimplifyTest.elm
```diff
@@ -7248,6 +7343,54 @@ a = Set.size (Set.fromList [b, c, d])
...
+                            , under = "Set.size"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = 2
+"""
+                        ]
+        , test "should replace Set.size (Set.fromList [([1, 2], [3, 4]), ([1, 2], [3, 4]), ([], [1])]) by 2" <|
+            \() ->
+                """module A exposing (..)
+a = Set.size (Set.fromList [([1, 2], [3, 4]), ([1, 2], [3, 4]), ([], [1])])
```
jfmengels, line 7383: Nice one :grin:

jfmengels (commented):
Thank you for working on this :pray: 

I think this looks good enough to merge. I have left some comments. I'd like for this check to also look for Strings if possible.

Let me know if you want to and have time to address them. If you don't, I'll just merge this (almost) as is probably :)

src/Simplify.elm
jiegillet: Strings and chars are supported, they are in the `_`. I can afford to do that because I know at this point that they are all comparable values. Now I see it would have been clearer to list them explicitly. 

To avoid iterating twice, I think I could re-implement `Maybe.Extra.traverse : (a -> Maybe b) -> List a -> Maybe (List b)`. I'll do that :)

jfmengels:
Thank you very much @jiegillet :pray: 
I just released your work in 2.0.11.

jiegillet:
Awesome, thank you :)

PR #35, opened 2022-08-19
Simplify field access for records and record updates

### LINKED ISSUE #34: Simplify field access to records and record updates
**What the rule should do:**

**What problems does it solve:**
Remove redundant record creation/update if immediately followed by a field access

**Example of things the rule would report:**

```elm
{ foo | a = 3 }.a
--> (3)

{ foo | a = f x y }.a
--> (f x y)

{ a = 3 }.a
--> (3)
```

**Example of things the rule would not report:**

```elm
{ foo | b = 3 }.a

{ b = 3 }.a 
```

Part of the Simplify rule.

**I am looking for:**
- Feedback
- Making a note using an Issue before I forget about this :D I actually plan on writing the PR myself

miniBill:
Closes #34

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1352,11 +1352,7 @@ expressionVisitorHelp node context =
                 _ ->
                     onlyErrors []
 
-        Expression.RecordAccess record field ->
-            let
-                fieldName =
-                    Node.value field
-            in
+        Expression.RecordAccess record (Node _ fieldName) ->
             case Node.value record of
                 Expression.RecordExpr setters ->
                     onlyErrors (recordAccessChecks (Node.range node) fieldName setters)
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1359,10 +1359,10 @@ expressionVisitorHelp node context =
             in
             case Node.value record of
                 Expression.RecordExpr setters ->
-                    onlyErrors (recordAccessChecks node fieldName setters)
+                    onlyErrors (recordAccessChecks (Node.range node) fieldName setters)
 
                 Expression.RecordUpdateExpression _ setters ->
-                    onlyErrors (recordAccessChecks node fieldName setters)
+                    onlyErrors (recordAccessChecks (Node.range node) fieldName setters)
 
                 _ ->
                     onlyErrors []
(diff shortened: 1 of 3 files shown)


src/Simplify.elm
```diff
@@ -1352,10 +1353,64 @@ expressionVisitorHelp node context =
...
+                    in
+                    ( Node.value nfield, nvalue )
+                )
+                setters
+
+        nodeRange =
+            Node.range node
+    in
+    case
+        List.Extra.find (\( setterField, _ ) -> setterField == fieldName) setterValues
```
jfmengels, line 1393: There is already a `find` function defined in the same module. No need for the dependency.

src/Simplify.elm
```diff
@@ -1352,10 +1353,64 @@ expressionVisitorHelp node context =
...
+            let
+                fieldName =
+                    Node.value field
+            in
+            case Node.value record of
+                Expression.RecordExpr setters ->
+                    onlyErrors (recordAccessChecks node fieldName setters)
+
+                Expression.RecordUpdateExpression _ setters ->
+                    onlyErrors (recordAccessChecks node fieldName setters)
```
jfmengels, line 1366: You don't need the entire node here, only it's range
```suggestion
                    onlyErrors (recordAccessChecks (Node.range node) fieldName setters)

                Expression.RecordUpdateExpression _ setters ->
                    onlyErrors (recordAccessChecks (Node.range node) fieldName setters)
```

jfmengels (changes requested):
This looks very good. I added a few remarks, but I don't think it's too much work.

src/Simplify.elm
miniBill: Ok, let me rewrite it

src/Simplify.elm
miniBill: Fair enough!

src/Simplify.elm
```diff
@@ -1352,10 +1353,64 @@ expressionVisitorHelp node context =
...
+recordAccessChecks node fieldName setters =
+    let
+        setterValues =
+            List.map
+                (\setter ->
+                    let
+                        ( nfield, nvalue ) =
+                            Node.value setter
+                    in
+                    ( Node.value nfield, nvalue )
```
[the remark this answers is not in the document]
miniBill: Ooh nice, didn't check whether the constructor was exposed

src/Simplify.elm
```diff
@@ -1352,10 +1353,64 @@ expressionVisitorHelp node context =
...
+            Node.range node
+    in
+    case
+        List.Extra.find (\( setterField, _ ) -> setterField == fieldName) setterValues
+    of
+        Just ( _, setterValue ) ->
+            let
+                setterRange =
+                    Node.range setterValue
+            in
```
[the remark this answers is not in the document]
miniBill: Oh yea, absolutely

jfmengels (approved):
LGTM, thanks a lot for the new simplification! Very nice addition!

PR #37, opened 2022-08-19
Simplify record access to unrelated field update

miniBill:


jfmengels:
Woohoo :partying_face: 

This one will probably apply more often than #35

Issue #33, opened 2022-08-19
Distribute field access across `ifThenElse`, `case` and `letIn`


miniBill:
**What the rule should do:**
Distribute field access across `ifThenElse`, `case` and `letIn`.

**What problems does it solve:**
It makes the code easier to read and unlocks other possible simplifications

**Example of things the rule would report:**

```elm
(if x then y else z).a
--> if x then y.a else z.a

(let x = y in z).a
--> let x = y in z.a

(case x of Foo y -> z \n Bar q -> w).a
--> case x of Foo y -> z.a \n Bar q -> w.a
```

**Should this be part of the Simplify rule or should it be a new rule?**
I am not sure

**I am looking for:**

- Feedback: when to apply it
  - Always
  - Only when all branches are "simple": either records, variables or record updates
  - Only when at least one branch is "simple" (as above)
- Tips and help on how to implement it
  - How to I maintain the code style while inserting parentheses and code
- Should this be part of Simplify or another rule?

PR #38, opened 2022-08-19
Add simplification for field access outside let/in expressions

miniBill:
Related to #33 

This only solves the `let/in` case which is more visibly correct

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1368,14 +1368,14 @@ expressionVisitorHelp node context =
                     onlyErrors []
 
         Expression.RecordAccess record (Node _ fieldName) ->
-            case Node.value record of
+            case Node.value (AstHelpers.removeParens record) of
                 Expression.RecordExpr setters ->
                     onlyErrors (recordAccessChecks (Node.range node) Nothing fieldName setters)
 
                 Expression.RecordUpdateExpression (Node recordNameRange _) setters ->
                     onlyErrors (recordAccessChecks (Node.range node) (Just recordNameRange) fieldName setters)
 
-                Expression.ParenthesizedExpression (Node _ (Expression.LetExpression { expression })) ->
+                Expression.LetExpression { expression } ->
                     onlyErrors (recordAccessLetInChecks (Node.range node) (Node.range expression))
 
                 _ ->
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -1372,6 +1375,9 @@ expressionVisitorHelp node context =
                 Expression.RecordUpdateExpression (Node recordNameRange _) setters ->
                     onlyErrors (recordAccessChecks (Node.range node) (Just recordNameRange) fieldName setters)
 
+                Expression.ParenthesizedExpression (Node _ (Expression.LetExpression { expression })) ->
+                    onlyErrors (recordAccessLetInChecks (Node.range node) (Node.range expression))
+
```
jfmengels, line 1380: Instead of doing parens here, you can also do `case Node.value (AstHelpers.removeParens record) of` a few lines above. I do think we should then add a few tests for the other cases to see whether they support parens ok.

Using this function can be nicer because we can then pile on more simplifications in `--fix-all` before `elm-format` kicks in (and asks one less prompt to the user in this situation).

But if this makes the fix a lot more annoying, let me know and let's skip it. Do you mind trying this out?

src/Simplify.elm
```diff
@@ -1372,6 +1375,9 @@ expressionVisitorHelp node context =
                 Expression.RecordUpdateExpression (Node recordNameRange _) setters ->
                     onlyErrors (recordAccessChecks (Node.range node) (Just recordNameRange) fieldName setters)
 
+                Expression.ParenthesizedExpression (Node _ (Expression.LetExpression { expression })) ->
```
jfmengels, line 1378: Instead of doing parens here, you can also do `case Node.value (AstHelpers.removeParens record) of` a few lines above. I do think we should then add a few tests for the other cases to see whether they support parens ok.

Using this function can be nicer because we can then pile on more simplifications in `--fix-all` before `elm-format` kicks in (and asks one less prompt to the user in this situation).

But if this makes the fix a lot more annoying, let me know and let's skip it. Do you mind trying this out?

src/Simplify.elm
miniBill: Adding that breaks another test, but it's probably a good break, checking what's up

jfmengels (approved):
Looks great. Thanks again @miniBill :blush: :heart:

PR #40, opened 2022-08-19
Add field access simplification for if/then/else

### LINKED ISSUE #33: Distribute field access across `ifThenElse`, `case` and `letIn`

miniBill:
Partially fixes #33

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1367,7 +1367,7 @@ expressionVisitorHelp node context =
                 _ ->
                     onlyErrors []
 
-        Expression.RecordAccess record ((Node _ fieldName) as field) ->
+        Expression.RecordAccess record field ->
             case Node.value (AstHelpers.removeParens record) of
                 Expression.RecordExpr setters ->
                     onlyErrors (recordAccessChecks (Node.range node) Nothing (Node.value field) setters)
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1367,7 +1367,7 @@ expressionVisitorHelp node context =
                 _ ->
                     onlyErrors []
 
-        Expression.RecordAccess record field ->
+        Expression.RecordAccess record ((Node _ fieldName) as field) ->
             case Node.value (AstHelpers.removeParens record) of
                 Expression.RecordExpr setters ->
                     onlyErrors (recordAccessChecks (Node.range node) Nothing (Node.value field) setters)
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -1378,13 +1378,63 @@ expressionVisitorHelp node context =
...
 
         _ ->
             onlyErrors []
 
 
+leavesAreRecords : Expression -> Bool
+leavesAreRecords expr =
+    case expr of
+        Expression.IfBlock _ (Node _ ifValue) (Node _ elseValue) ->
+            leavesAreRecords ifValue && leavesAreRecords elseValue
```
jfmengels, line 1418: The current implementation will figure out that long ifthenelse chains have records in the leaves, but then in the fix you only replace the root then and else, which will be somewhat wasteful.

I don't think we want this recursive pattern here (for the if/else), at least until we decide we want to apply this transformation to long chains of if/else.

src/Simplify.elm
```diff
@@ -1378,13 +1378,63 @@ expressionVisitorHelp node context =
                 Expression.LetExpression { expression } ->
                     onlyErrors (recordAccessLetInChecks (Node.range node) (Node.range expression))
 
+                Expression.IfBlock _ (Node thenRange thenValue) (Node elseRange elseValue) ->
+                    let
+                        nodeRange : Range
+                        nodeRange =
+                            Node.range node
+                    in
+                    if leavesAreRecords thenValue && leavesAreRecords elseValue then
```
jfmengels, line 1387: One thing that you can do is to, instead of returning a boolean in `leavesAreRecords`, return the ranges of the found records, maybe in the shape of `Maybe (List (Node Range))`, returning `Nothing` when one the leaves isn't a record.

That way, you can then use those records to insert the record access

```suggestion
                    case leavesAreRecords [ thenValue, elseValue ] 
                      Just ranges ->
                        onlyErrors [ Rule.error ...
                           ( Fix.removeRange { start = (Node.range record).end, end = nodeRange.end } :: (List.concatMap <insert record access> ranges))
```

(rough draft, clean it up as you see fit)

tests/SimplifyTest.elm
```diff
@@ -10424,4 +10424,27 @@ a = (let b = c in (f x).e).f
...
+"""
+                    |> Review.Test.run (rule defaults)
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Field access can be simplified"
+                            , details = [ "Accessing the field outside an if/then/else expression can be simplified to access the field inside it" ]
+                            , under = "(if x then { f = 3 } else { z | f = 3 }).f"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = (if x then ({ f = 3 }).f else ({ z | f = 3 }).f)
```
jfmengels, line 10440: Let's make an improvement in a future PR (I can do it if you want, but feel free to work on it) to not add parentheses unnecessarily. That will make for better diffs and less unnecessary fixes.

tests/SimplifyTest.elm
```diff
@@ -10424,4 +10424,27 @@ a = (let b = c in (f x).e).f
...
+            \() ->
+                """module A exposing (..)
+a = (if x then { f = 3 } else { z | f = 3 }).f
+"""
+                    |> Review.Test.run (rule defaults)
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Field access can be simplified"
+                            , details = [ "Accessing the field outside an if/then/else expression can be simplified to access the field inside it" ]
+                            , under = "(if x then { f = 3 } else { z | f = 3 }).f"
```
jfmengels, line 10437: The under is getting really big here. In a real example, this would put squiggly lines over multiple lines (since ifs are always on multiple lines), which can be very distracting and annoying to a user.

I think that we should change the "under" to only be the `.f` in all of the simplifications we've done around record access. What do you think?

jfmengels (changes requested):

src/Simplify.elm
miniBill: I definitely, intentionally, want this applied to long chains. If all the leaves are records then it will simplify all of them.

src/Simplify.elm
miniBill: Oh, yes, that makes a lot of sense, so we don't require multiple passes!

tests/SimplifyTest.elm
miniBill: I'd prefer it if you worked on it, to be honest

tests/SimplifyTest.elm
miniBill: Agreed!

tests/SimplifyTest.elm
jfmengels: That's fine, I'll do that at some point 👍

miniBill: I've actually achieved most of the no-redundant-parens result ^_^

miniBill:
If merged this would close #33

src/Simplify.elm
```diff
@@ -1385,6 +1391,93 @@ expressionVisitorHelp node context =
...
+distributeFieldAccess : String -> Node Expression -> Node String -> List (Error {})
+distributeFieldAccess kind ((Node recordRange _) as record) (Node fieldRange fieldName) =
+    case recordLeavesRanges record of
+        { records, withoutParens, withParens } ->
+            if List.length withParens > 1 then
+                -- Never introduce more than one set of parentheses
+                -- This is an heuristic for avoiding distributing field access
+                -- when the expressions are not amenable to simplification
+                -- while still letting it run when there is a single exception
+                -- or if it's a let/in
```
jfmengels, line 1403: I think it would be better if following what Simon Lydell said: Only apply the transformation if we see at least one place where we see a simplification to apply. I think we could do the same checks that we do in `recordAccessChecks`, and if we have at least one branch where it matches, then we report an error, otherwise we don't.

That sounds like a more useful approach than this heuristic, where people will have trouble figuring out why the optimization didn't kick in.

miniBill: So `if List.isEmpty records`? Makes sense. And then a boolean to do it anyway for `let`s?

miniBill: Pushed

miniBill:
Rebased

jfmengels:
Hey @miniBill I'm not sure what the state of this PR is. I don't think the direction this is heading towards is the one. I think that this would make sense if we notice that in every branch, we can end up simplifying the pattern, but not otherwise. What do you think?

miniBill:
It seems like a reasonable restriction! It's uniform, non-surprising, and should have very few false positives!

miniBill:
Ok, I've restricted the simplification to the case where all leaves are records, record updates or record accesses

tests/SimplifyTest.elm
```diff
@@ -10443,49 +10443,49 @@ a = d.c
         , test "should simplify record accesses for let/in expressions" <|
             \() ->
                 """module A exposing (..)
-a = (let b = c in f x).e
+a = (let b = c in { e = 3 }).e
```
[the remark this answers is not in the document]
miniBill: Ok, done ^_^
Now it applies if it's a let or if it's only "record-like" expressions (records, updates, field accesses)

jfmengels (approved):

jfmengels:
Looks good, thank you very much for your work and patience @miniBill :pray: :blush:

miniBill:
Happy to have it merged ^_^

PR #44, opened 2022-08-22
Take ranges into consideration when doing inference on numbers

miniBill:
**WIP**

This doesn't break existing tests, but it doesn't fix the new one either

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1367,7 +1367,7 @@ expressionVisitorHelp node context =
                 _ ->
                     onlyErrors []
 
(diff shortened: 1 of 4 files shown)


miniBill:
This is going to be transformed into elm-review-optimize

jfmengels:
Sounds good. Please share your journey with me, I'd love to be give advice and pitch in the idea! :slightly_smiling_face:

miniBill:
https://github.com/miniBill/elm-review-optimize current status is: I broke two tests and I'm working on fixing them. Then I'll write more tests and then I'll remove things that elm-review-simplify already does

PR #45, opened 2022-08-26
Don't simplify case expressions containing custom type constructors from the project

### LINKED ISSUE #36: Add type path to unneccesary case error
## Pain point

When adding `elm-review-simplify` to a project that is reasonably mature already, it seems very likely to run into `Unneccesary case expression` errors:

```
(fix) Simplify: Unnecessary case expression

128| viewCheckoutError findOfferError =
129|     case findOfferError of
         ^^^^
130|         Offer.NotInValidityPeriodError _ _ ->

All the branches of this case expression resolve to the same value. You can
remove the case expression and replace it with the body of one of the branches.
```

If I don't want to tackle these right now, I might reference these types in an `Simplify.ignoreCaseOfForTypes` call.

Right now I have to manually go into the codebase and track down the module path leading to this type which is a bit tedious.

## possible solution

I think it would be convenient if the rule already did this work for me by adding to the error message:

```
(fix) Simplify: Unnecessary case expression

128| viewCheckoutError findOfferError =
129|     case findOfferError of
         ^^^^
130|         Offer.NotInValidityPeriodError _ _ ->

All the branches of this case expression resolve to the same value. You can
remove the case expression and replace it with the body of one of the branches.

If this was intentional you can surpress this warning 
by adding `Path.To.Type` to the rule configuration through `Simplify.ignoreCaseOfForTypes`. 

I personally don't recommend to use this function too much, 
because this could be a sign of premature abstraction, 
and because I think that often [You Aren't Gonna Need this code](https://jfmengels.net/safe-dead-code-removal/#yagni-you-arent-gonna-need-it).
```

jfmengels:
Fixes #36 

The rule now doesn't simplify case expressions that refer to a custom type defined in the project, though it will keep doing this for custom types that come from dependencies (`Just _ -> ...`) as well as patterns like `1 ->`, `[] ->`, etc.

@janiczek @stefan-wullems @lydell if you'd be willing to try this change and tell me what the experience is like, that would be great :pray: 

If you configured the rule to use `ignoreCaseOfForTypes`, you will likely get new global errors.

```
elm-review --template jfmengels/elm-review-simplify/preview#case-of-constructors
```


lydell:
The status at work is:

- We want to use the simplify rule, but it wanting to simplify many case-ofs has made us not enable the rule, because there are too many places where it’s unclear how we should solve it. It seems to be 15 places, which right now doesn’t seem like that much, but apparently enough for us to never having enabled the rule.
- I ran this PR one the code base. It’s perfect! All the simplifications suggested were relevant and good, and it still found a `case foo of Just _ -> bar Nothing -> bar` case and simplified it.
- It found zero places where we cased on other custom types from dependencies unnecessarily. All of our “unnecessary” case-ofs seem to be of types from our own code.
- I don’t have use for the option to exclude some custom types.

I’m excited! Apart from simplifying some things, the rule also found some conditions like `questionId == questionId &&` which I need to take a look at! With this change we can enable this rule at work and get these benefits all the time.

PR #47, opened 2022-09-05
Simplify record access simplifications

jfmengels:
This will in my opinion be a bit easier to understand.
For the records part, it should also be slightly faster because there is less data creation, as well as shortcutting where we stop the exploration as soon as we see something that will prevent the simplification (and some tail-call optimization)

cc @miniBill if you'd like to take a look

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1218,10 +1218,10 @@ expressionVisitorHelp node context =
                     onlyErrors [ injectRecordAccessIntoLetExpression "a let/in" (Node.range record) expression field ]
 
                 Expression.IfBlock _ _ _ ->
-                    onlyErrors (distributeFieldAccess False "an if/then/else" record field)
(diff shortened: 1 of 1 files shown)


miniBill:
Inverse condition: oh yeah, absolutely, I mostly didn't do it to avoid noise.
Use a recursive approach: oooh, that's smart! Love it!
Use let destructuring: I can't remember why I had a `case` there, but yeah, that `let` is simpler
Explicit case for let expression/remove unnecessary parameter: Yeah, now that the rules are very different it makes sense to split again
No need to handle parens and without parens: And to remove this
Use shortcutting recursive version: Yupp
Rename: Makes sense
Pass a list of Nodes: Is this simply to avoid destructuring twice? If so, makes sense, otherwise I'm missing something
Remove constant parameters: _nodnod_

jfmengels:
> Pass a list of Nodes
Yes, it's mostly so we don't pattern match on "record" (and the version with parentheses) twice.

PR #49, opened 2022-09-06
Fix bad link in README

bdukes:


jfmengels:
Thanks for reporting the issue @bdukes :pray: 

I fixed the issue myself because the correct fix was *slightly* different:
- The closing parens needs to be after "Provides automatic fixes" (so that that becomes the alt text)
- The link should now point to v2.0.21, not v2.0.19

The bug was introduced by a faulty tool, which I will now update so that this doesn't happy anymore :)

PR #55, opened 2023-01-07
`String` operation simplifications

lue-bird:
Adds `String` simplifications (and tests) suggested in #2
```elm
String.slice n n string
--> ""

String.slice 0 n string
--> String.left n string

String.slice n 0 string
--> ""

String.slice a z ""
--> ""

String.left 0 string
--> ""

String.left -1 string
--> ""

String.left n ""
--> ""

String.right 0 string
--> ""

String.right -1 string
--> ""

String.right n ""
--> ""

String.slice 1 2 "abc"
--> "b"

String.left 2 "abc"
--> "ab"

String.right 2 "abc"
--> "bc"
```
There rest is minor changes like "a empty" → "an empty"

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -2908,6 +2908,15 @@ stringReverseChecks ({ parentRange, fnRange, firstArg } as checkInfo) =
 stringSliceChecks : CheckInfo -> List (Error {})
 stringSliceChecks checkInfo =
     case ( checkInfo.firstArg, checkInfo.secondArg, checkInfo.thirdArg ) of
+        ( _, Just (Node _ (Expression.Integer 0)), _ ) ->
+            [ Rule.errorWithFix
+                { message = "Using String.slice with end index 0 will result in an empty string"
+                , details = [ "You can replace this call by an empty string." ]
+                }
+                checkInfo.fnRange
+                [ Fix.replaceRangeBy checkInfo.parentRange "\"\"" ]
+            ]
+
         ( Node _ (Expression.Integer 0), _, _ ) ->
             [ Rule.errorWithFix
                 { message = "Use String.left instead"
@@ -2920,15 +2929,6 @@ stringSliceChecks checkInfo =
                     }
                     "String.left"
                 ]
-            ]
-
-        ( _, Just (Node _ (Expression.Integer 0)), _ ) ->
-            [ Rule.errorWithFix
-                { message = "Using String.slice with end index 0 will result in an empty string"
-                , details = [ "You can replace this call by an empty string." ]
-                }
-                checkInfo.fnRange
-                [ Fix.replaceRangeBy checkInfo.parentRange "\"\"" ]
             ]
 
         ( _, _, Just (Node _ (Expression.Literal "")) ) ->
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -2907,31 +2907,8 @@ stringReverseChecks ({ parentRange, fnRange, firstArg } as checkInfo) =
 
 stringSliceChecks : CheckInfo -> List (Error {})
 stringSliceChecks checkInfo =
-    case ( checkInfo.firstArg, checkInfo.secondArg, checkInfo.thirdArg ) of
-        ( _, Just (Node _ (Expression.Integer 0)), _ ) ->
-            [ Rule.errorWithFix
-                { message = "Using String.slice with end index 0 will result in an empty string"
-                , details = [ "You can replace this call by an empty string." ]
-                }
-                checkInfo.fnRange
-                [ Fix.replaceRangeBy checkInfo.parentRange "always \"\"" ]
-            ]
-
-        ( Node _ (Expression.Integer 0), _, _ ) ->
-            [ Rule.errorWithFix
-                { message = "Use String.left instead"
-                , details = [ "Using String.slice with start index 0 is the same as using String.left." ]
-                }
-                checkInfo.fnRange
-                [ Fix.replaceRangeBy
-                    { start = checkInfo.fnRange.start
-                    , end = (Node.range checkInfo.firstArg).end
-                    }
-                    "String.left"
-                ]
-            ]
-
-        ( _, _, Just (Node _ (Expression.Literal "")) ) ->
+    case checkInfo.thirdArg of
+        Just (Node _ (Expression.Literal "")) ->
             [ Rule.errorWithFix
                 { message = "Using String.slice on an empty string will result in an empty string"
                 , details = [ "You can replace this call by an empty string." ]
@@ -2940,21 +2917,46 @@ stringSliceChecks checkInfo =
                 [ Fix.replaceRangeBy checkInfo.parentRange "\"\"" ]
             ]
 
-        ( start, Just end, _ ) ->
-            if Normalize.areAllTheSame checkInfo start [ end ] then
-                [ Rule.errorWithFix
-                    { message = "Using String.slice with equal start and end index will result in an empty string"
-                    , details = [ "You can replace this call by an empty string." ]
-                    }
-                    checkInfo.fnRange
-                    [ Fix.replaceRangeBy checkInfo.parentRange "always \"\"" ]
-                ]
+        _ ->
+            case ( checkInfo.firstArg, checkInfo.secondArg ) of
+                ( _, Just (Node _ (Expression.Integer 0)) ) ->
+                    [ Rule.errorWithFix
+                        { message = "Using String.slice with end index 0 will result in an empty string"
+                        , details = [ "You can replace this call by an empty string." ]
+                        }
+                        checkInfo.fnRange
+                        [ Fix.replaceRangeBy checkInfo.parentRange "always \"\"" ]
+                    ]
 
-            else
-                []
+                ( Node _ (Expression.Integer 0), _ ) ->
+                    [ Rule.errorWithFix
+                        { message = "Use String.left instead"
+                        , details = [ "Using String.slice with start index 0 is the same as using String.left." ]
+                        }
+                        checkInfo.fnRange
+                        [ Fix.replaceRangeBy
+                            { start = checkInfo.fnRange.start
+                            , end = (Node.range checkInfo.firstArg).end
+                            }
+                            "String.left"
+                        ]
+                    ]
 
-        ( _, Nothing, _ ) ->
-            []
+                ( start, Just end ) ->
+                    if Normalize.areAllTheSame checkInfo start [ end ] then
+                        [ Rule.errorWithFix
+                            { message = "Using String.slice with equal start and end index will result in an empty string"
+                            , details = [ "You can replace this call by an empty string." ]
+                            }
+                            checkInfo.fnRange
+                            [ Fix.replaceRangeBy checkInfo.parentRange "always \"\"" ]
+                        ]
+
+                    else
+                        []
+
+                ( _, Nothing ) ->
+                    []
 
 
 stringLeftChecks : CheckInfo -> List (Error {})
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -3012,7 +3012,7 @@ stringRightChecks checkInfo =
                 [ Fix.replaceRangeBy checkInfo.parentRange "\"\"" ]
             ]
 
-        ( Node _ (Expression.Negation (Node _ (Expression.Integer length))), _ ) ->
+        ( Node _ (Expression.Negation (Node _ (Expression.Integer _))), _ ) ->
             [ Rule.errorWithFix
                 { message = "Using String.right with negative length will result in an empty string"
                 , details = [ "You can replace this call by an empty string." ]
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -2614,18 +2614,18 @@ basicsIdentityChecks checkInfo =
     ]
 
 
+identityCompositionErrorMessage : { message : String, details : List String }
+identityCompositionErrorMessage =
+    { message = "`identity` should be removed"
+    , details = [ "Composing a function with `identity` is the same as simplify referencing the function." ]
+    }
+
+
 identityCompositionCheck : CompositionCheckInfo -> List (Error {})
 identityCompositionCheck { lookupTable, left, right } =
-    let
-        errorInfo : () -> { message : String, details : List String }
-        errorInfo () =
-            { message = "`identity` should be removed"
-            , details = [ "Composing a function with `identity` is the same as simplify referencing the function." ]
-            }
-    in
     if isIdentity lookupTable right then
         [ Rule.errorWithFix
-            (errorInfo ())
+            identityCompositionErrorMessage
             (Node.range right)
             [ Fix.removeRange { start = (Node.range left).end, end = (Node.range right).end }
             ]
@@ -2633,7 +2633,7 @@ identityCompositionCheck { lookupTable, left, right } =
 
     else if isIdentity lookupTable left then
         [ Rule.errorWithFix
-            (errorInfo ())
+            identityCompositionErrorMessage
             (Node.range left)
             [ Fix.removeRange { start = (Node.range left).start, end = (Node.range right).start }
             ]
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -2876,6 +2905,126 @@ stringReverseChecks ({ parentRange, fnRange, firstArg } as checkInfo) =
...
+            [ Rule.errorWithFix
+                { message = "Using String.slice on an empty string will result in an empty string"
+                , details = [ "You can replace this call by an empty string." ]
+                }
+                checkInfo.fnRange
+                [ Fix.replaceRangeBy checkInfo.parentRange "\"\"" ]
+            ]
+
+        ( start, Just end, _ ) ->
+            if Normalize.areAllTheSame checkInfo start [ end ] then
```
lue-bird, line 2944: This one I'm not fully sure about. We want to check if both arguments are equivalent expressions... Is "normalizing" what we are looking for here?

jfmengels: Yes, that sounds good :+1:

src/Simplify.elm
```diff
@@ -2872,6 +2873,49 @@ stringReverseChecks ({ parentRange, fnRange, firstArg } as checkInfo) =
                 checkInfo
 
 
+stringSliceChecks : CheckInfo -> List (Error {})
+stringSliceChecks checkInfo =
+    case ( checkInfo.firstArg, checkInfo.secondArg ) of
+        ( Node _ (Expression.Integer 0), _ ) ->
```
jfmengels: I think the order of checks could be improved, so that `String.slice 0 0 str` does not end up being `String.left 0 str`. You are adding support for this in this PR, but by changing the order we could avoid an additional round-trip.

tests/SimplifyTest.elm
```diff
@@ -5062,6 +5063,99 @@ a = x
...
+"""
+                    |> Review.Test.run (rule defaults)
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Using String.slice with end index 0 will result in an empty string"
+                            , details = [ "You can replace this call by an empty string." ]
+                            , under = "String.slice"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = ""
```
jfmengels, line 5091: This is incorrect: You can only do this when all 3 arguments have been passed to `String.slice`.

```elm
String.slice b 0
--> `always ""`
```
(Not sure whether parentheses are needed here)

tests/SimplifyTest.elm
```diff
@@ -5062,6 +5063,99 @@ a = x
...
```
jfmengels, line 5107: This one is correct though :+1:

tests/SimplifyTest.elm
```diff
@@ -5062,6 +5063,99 @@ a = x
...
+"""
+                        ]
+        , test "should replace String.slice n n str by \"\"" <|
+            \() ->
+                """module A exposing (..)
+a = String.slice n n str
```
jfmengels, line 5113: We should have a test for `a = String.slice n n`, which should be fixed to `a = always ""` like what I mentioned above.

tests/SimplifyTest.elm
```diff
@@ -5062,6 +5063,99 @@ a = x
...
+                            , under = "String.slice"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = ""
+"""
+                        ]
+        , test """should replace String.slice 0 n str by String.left n str""" <|
+            \() ->
+                """module A exposing (..)
+a = String.slice 0 n str
```
jfmengels, line 5145: Again, would love a test with less arguments.

```
String.slice 0 n str
--> String.left n str -- this one you already have

String.slice 0 n
--> String.left n

String.slice 0
--> String.left
```

tests/SimplifyTest.elm
```diff
@@ -5062,6 +5063,99 @@ a = x
...
+                            , under = "String.slice"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = String.left n str
+"""
+                        ]
+        , test """should replace str |> String.slice 0 n by str |> String.left n""" <|
+            \() ->
+                """module A exposing (..)
+a = str |> String.slice 0 n
```
jfmengels, line 5161: A test for `n |> String.slice 0` could be good too. And for `String.slice 0 n <| str` and `String.slice 0 <| n`.

Yeah sorry, I wish I could find an easy way to test them all in one go :thinking:

src/Simplify.elm
```diff
@@ -2875,8 +2875,8 @@ stringReverseChecks ({ parentRange, fnRange, firstArg } as checkInfo) =
 
 stringSliceChecks : CheckInfo -> List (Error {})
 stringSliceChecks checkInfo =
-    case ( checkInfo.firstArg, checkInfo.secondArg ) of
-        ( Node _ (Expression.Integer 0), _ ) ->
+    case ( checkInfo.firstArg, checkInfo.secondArg, checkInfo.thirdArg ) of
```
jfmengels, line 2910: I think nested if expressions could make more sense for readbility here. What do you think? :thinking:

tests/SimplifyTest.elm
```diff
@@ -5172,6 +5174,128 @@ a = str |> String.left n
...
+stringLeftTests =
+    describe "String.left"
+        [ test "should not report String.left that contains variables or expressions" <|
+            \() ->
+                """module A exposing (..)
+a = String.left b c
+"""
+                    |> Review.Test.run (rule defaults)
+                    |> Review.Test.expectNoErrors
+        , test """should replace String.left 0 by \"\"""" <|
```
jfmengels, line 5187: I think the triple quotes in the test titles are not helping, since you're escaping the quotes inside the title anyway :sweat_smile:

tests/SimplifyTest.elm
```diff
@@ -5172,6 +5174,128 @@ a = str |> String.left n
...
+"""
+                    |> Review.Test.run (rule defaults)
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Using String.left with length 0 will result in an empty string"
+                            , details = [ "You can replace this call by an empty string." ]
+                            , under = "String.left"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = ""
```
jfmengels, line 5200: Incorrect, should be `always ""`

tests/SimplifyTest.elm
```diff
@@ -5172,6 +5174,128 @@ a = str |> String.left n
...
+"""
+                    |> Review.Test.run (rule defaults)
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Using String.left with negative length will result in an empty string"
+                            , details = [ "You can replace this call by an empty string." ]
+                            , under = "String.left"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = ""
```
jfmengels, line 5216: Incorrect, should be always ""

tests/SimplifyTest.elm
```diff
@@ -5172,6 +5174,128 @@ a = str |> String.left n
...
+"""
+                    |> Review.Test.run (rule defaults)
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Using String.right with length 0 will result in an empty string"
+                            , details = [ "You can replace this call by an empty string." ]
+                            , under = "String.right"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = ""
```
jfmengels, line 5261: Incorrect, should be always ""

tests/SimplifyTest.elm
```diff
@@ -5172,6 +5174,128 @@ a = str |> String.left n
...
+"""
+                    |> Review.Test.run (rule defaults)
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Using String.right with negative length will result in an empty string"
+                            , details = [ "You can replace this call by an empty string." ]
+                            , under = "String.right"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = ""
```
jfmengels, line 5277: Incorrect, should be always ""

src/Simplify.elm
```diff
@@ -300,7 +300,7 @@ Destructuring using case expressions
     --> x
 
 
-### Maybe
+### Maybes
```
jfmengels, line 333: Good catch :+1:

tests/SimplifyTest.elm
lue-bird: right. Those came from copying concat/join/repeat tests that had the same inconsistency. Changed those, too

src/Simplify.elm
lue-bird: Same up there with "field access" header. Tho I don't have a good plural for "accesses"

src/Simplify.elm
lue-bird: I think either option looks a bit cursed... like
```elm
if Maybe.map Node.value checkInfo.thirdArg == Just (Expression.Literal "") then
```

lue-bird:
Yes, let's add 
```elm
String.slice 1 2 "abc" --> "b"
String.left 2 "abc" --> "ab"
String.right 2 "abc" --> "bc"
```

src/Simplify.elm
jfmengels: Right. I think that the condition is not great (though not that bad, I'm fine with it), but if it makes the rest of the conditions easier, I think it's probably better. Feel free to leave this out if you want to, I can take a look later.

src/Simplify.elm
jfmengels: I think "accesses" is the correct plural. But "Field access" is fine also. Leave that one as is :man_shrugging:

jfmengels:
Okay. I'm a bit more wary about these ones but I think it's okay to have them. Do beware of not messing things with Unicode characters though :/

Also, could you add a test for `String.slice 0 0 n`?

lue-bird:
> Do beware of not messing things with Unicode characters though

bit confused... How can that be possible? Something to do with how elm-syntax parses things into `Literal`?
```elm
case Maybe.map2 Tuple.pair (Evaluate.getInt checkInfo start) (Evaluate.getInt checkInfo end) of
    Nothing ->
        []

    Just ( startInt, endInt ) ->
        [ let
            sliced =
                String.slice startInt endInt nonEmptyString
            in
            Rule.errorWithFix
            { message = "The call to String.slice will result in \"" ++ sliced ++ "\""
            , details = [ "You can replace this slice operation by \"" ++ sliced ++ "\"." ]
            }
            checkInfo.fnRange
            [ Fix.replaceRangeBy
                checkInfo.parentRange
                sliced
            ]
        ]
```

jfmengels:
> bit confused... How can that be possible? Something to do with how elm-syntax parses things into Literal?

I don't know. And that's the problem :sweat_smile: 
I can imagine that there could be differences with unicode characters (where a character is in practice 2 characters) or with escaped characters (`\"` for instance). I'd love a few more tests to make sure we're not changing the result :/

lue-bird:
Well you were definitely correct (again). My naiive approach can't work because one of below is true
- `Literal` parsing doesn't un-escape, so writing in elm `"\u{000D}"` will give you an elm `String` `"\\u{000D}"`. Slicing this would be lethal
- `Literal` parsing un-escapes, slicing works as expected but we have to re-escape because the sliced `String`s might not be valid in elm (like " or control characters). That means figuring out which `Char`s are printable, returning hex, \ escaping, ... (https://github.com/avh4/elm-format/blob/main/elm-format-lib/src/ElmFormat/Render/Box.hs#L1948)

Not sure which of the 2 is what actually happens but both are a breaker in terms of effort etc.

I think it still makes sense to at least cover cases `identity` and `always ""` for `String` operations.
For `List.take`,`drop` it's a different picture, tho

jfmengels:
I think it makes sense to (at least in this PR, we can always revisit) not try to compute the result of `String.slice` based on the value of the string. Let's keep it at looking at the index values and checking whether the string is empty.

We can still have `String.slice 10 4 str` be simplified to `str`, since 10 < 4. But let's not evaluate `String.slice 4 10 "some literal"`.

lue-bird:
Yeah that'swhat I was trying to say in the last sentence :)

src/Simplify.elm
```diff
@@ -2876,6 +2911,154 @@ stringReverseChecks ({ parentRange, fnRange, firstArg } as checkInfo) =
...
+                [ Fix.replaceRangeBy checkInfo.parentRange "\"\"" ]
+            ]
+
+        ( _, Just (Node _ (Expression.Integer 0)), _ ) ->
+            [ Rule.errorWithFix
+                { message = "Using String.slice with end index 0 will result in an empty string"
+                , details = [ "You can replace this call by an empty string." ]
+                }
+                checkInfo.fnRange
+                [ Fix.replaceRangeBy checkInfo.parentRange "always \"\"" ]
```
jfmengels, line 2932: This should probably be `""` if there is a third argument, and ``always ""`` if there isn't, no?

src/Simplify.elm
```diff
@@ -2876,6 +2911,154 @@ stringReverseChecks ({ parentRange, fnRange, firstArg } as checkInfo) =
...
+                [ Fix.replaceRangeBy checkInfo.parentRange "\"\"" ]
+            ]
+
+        ( Node _ (Expression.Integer 0), _ ) ->
+            [ Rule.errorWithFix
+                { message = "Using String.left with length 0 will result in an empty string"
+                , details = [ "You can replace this call by an empty string." ]
+                }
+                checkInfo.fnRange
+                [ Fix.replaceRangeBy checkInfo.parentRange "always \"\"" ]
```
jfmengels, line 3012: This should probably be `""` if there is a second argument, and ``always ""`` if there isn't.

src/Simplify.elm
```diff
@@ -2876,6 +2911,154 @@ stringReverseChecks ({ parentRange, fnRange, firstArg } as checkInfo) =
...
+            ]
+
+        ( Node _ (Expression.Negation (Node _ (Expression.Integer _))), _ ) ->
+            [ Rule.errorWithFix
+                { message = "Using String.left with negative length will result in an empty string"
+                , details = [ "You can replace this call by an empty string." ]
+                }
+                checkInfo.fnRange
+                [ Fix.replaceRangeBy checkInfo.parentRange "always \"\"" ]
```
jfmengels, line 3021: This should probably be `""` if there is a second argument, and ``always ""`` if there isn't.

src/Simplify.elm
```diff
@@ -2876,6 +2911,154 @@ stringReverseChecks ({ parentRange, fnRange, firstArg } as checkInfo) =
...
+                [ Fix.replaceRangeBy checkInfo.parentRange "\"\"" ]
+            ]
+
+        ( Node _ (Expression.Integer 0), _ ) ->
+            [ Rule.errorWithFix
+                { message = "Using String.right with length 0 will result in an empty string"
+                , details = [ "You can replace this call by an empty string." ]
+                }
+                checkInfo.fnRange
+                [ Fix.replaceRangeBy checkInfo.parentRange "always \"\"" ]
```
jfmengels, line 3046: This should probably be `""` if there is a second argument, and ``always ""`` if there isn't.

src/Simplify.elm
```diff
@@ -2876,6 +2911,154 @@ stringReverseChecks ({ parentRange, fnRange, firstArg } as checkInfo) =
...
+            ]
+
+        ( Node _ (Expression.Negation (Node _ (Expression.Integer _))), _ ) ->
+            [ Rule.errorWithFix
+                { message = "Using String.right with negative length will result in an empty string"
+                , details = [ "You can replace this call by an empty string." ]
+                }
+                checkInfo.fnRange
+                [ Fix.replaceRangeBy checkInfo.parentRange "always \"\"" ]
```
jfmengels, line 3055: This should probably be `""` if there is a second argument, and ``always ""`` if there isn't.

jfmengels (changes requested):
I think this is looking good, and very close to merge. But the incorrect fixes need to be removed first.

src/Simplify.elm
lue-bird: no because the replaced range only ranged up to the second :)

lue-bird: Thanks! I thought I had updated the range to only go up until including the last argument we know of... guess not, let me fix

jfmengels (approved):
Alright, this looks good. I will make some minor edits in a separate PR, but this is good enough to merge :+1: 

Thank you for all these lovely simplifications @lue-bird :heart:

PR #56, opened 2023-01-10
Avoid using always when we can replace directly by the value

jfmengels:
The idea is to have
```elm
String.slice 1 1 str
```
replaced by
```elm
""
```
instead of by
```elm
always "" str
```
and the same for similar replacements done in #55. The rule would eventually end up being `""` but this removes some step and eventual confusion for a user running single fixes at a time.

cc @lue-bird if you'd like to review :smile: (no pressure!)


lue-bird:
Great idea and great abstraction. I think you forgot the 2 `always ""` in `stringRightChecks`

jfmengels:
Good catch. Fixed now :+1:

lue-bird (approved):

PR #57, opened 2023-01-12
`List.fold l,r` checks added

lue-bird:
Adds a couple `List.foldl` and `foldr` simplifications
```elm
List.foldl reduce initial []
--> initial

List.foldl 0 (+) list
--> List.sum list

List.foldl initial (+) list
--> initial + List.sum list

List.foldl 1 (*) list
--> List.product list

List.foldl initial (*) list
--> initial + List.product list

List.foldl True (&&) list
--> List.all identity list

List.foldl False (&&) list
--> False

List.foldl False (||) list
--> List.any identity list

List.foldl True (||) list
--> True

List.foldl fn x []
--> x

List.foldl (always identity) x list
--> x
```
same with `foldr`
```elm
List.foldr reduce initial []
--> initial

List.foldr 0 (+) list
--> List.sum list

List.foldr initial (+) list
--> initial + List.sum list

List.foldr 1 (*) list
--> List.product list

List.foldr initial (*) list
--> initial + List.product list

List.foldr True (&&) list
--> List.all identity list

List.foldr False (&&) list
--> False

List.foldr False (||) list
--> List.any identity list

List.foldr True (||) list
--> True

List.foldr fn x []
--> x

List.foldr (always identity) x list
--> x
```
There are places in the code marked TODO that we can discuss

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -513,6 +513,9 @@ Destructuring using case expressions
     List.foldl (*) 1 list
     --> List.product list
 
+    List.foldl (*) 0 list
+    --> 0
+
     List.foldl (*) initial list
     --> initial + List.product list
 
@@ -542,6 +545,9 @@ Destructuring using case expressions
 
     List.foldr (*) initial list
     --> initial + List.product list
+
+    List.foldr (*) 0 list
+    --> 0
 
     List.foldr (&&) True list
     --> List.all identity list
@@ -3860,7 +3866,18 @@ listFoldAnyDirectionChecks foldOperationName checkInfo =
                 numberBinaryOperationChecks { two = "+", list = "sum", identity = 0 }
 
             else if isBinaryOperation "*" checkInfo checkInfo.firstArg then
-                numberBinaryOperationChecks { two = "*", list = "product", identity = 1 }
+                if getUncomputedNumberValue initialArgument == Just 0 then
+                    [ Rule.errorWithFix
+                        { message = "The call to List." ++ foldOperationName ++ " (*) 0 will result in 0."
+                        , details = [ "You can replace this call by 0." ]
+                        }
+                        checkInfo.fnRange
+                        (replaceByEmptyFix "0" checkInfo.parentRange checkInfo.thirdArg)
+                    ]
+
+                else
+                    -- getUncomputedNumberValue initialArgument /= Just 0
+                    numberBinaryOperationChecks { two = "*", list = "product", identity = 1 }
 
             else if isBinaryOperation "&&" checkInfo checkInfo.firstArg then
                 boolBinaryOperationChecks { two = "&&", list = "all", determining = False }
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -3737,9 +3737,9 @@ listFoldAnyDirectionChecks foldOperationName checkInfo =
                 }
                 checkInfo.fnRange
                 [ Fix.removeRange
-                    { start = initialArgumentRange.end, end = checkInfo.parentRange.end }
-                , Fix.removeRange
                     { start = checkInfo.parentRange.start, end = initialArgumentRange.start }
+                , Fix.removeRange
+                    { start = initialArgumentRange.end, end = checkInfo.parentRange.end }
                 ]
             ]
 
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -507,11 +507,11 @@ Destructuring using case expressions
     List.foldl 0 (+) list
     --> List.sum list
 
-    List.foldl 1 (*) list
-    --> List.product list
-
     List.foldl initial (+) list
     --> initial + List.sum list
+
+    List.foldl 1 (*) list
+    --> List.product list
 
     List.foldl initial (*) list
     --> initial + List.product list
@@ -521,6 +521,12 @@ Destructuring using case expressions
 
     List.foldl False (&&) list
     --> False
+
+    List.foldl False (||) list
+    --> List.any identity list
+
+    List.foldl True (||) list
+    --> True
 
     List.foldl fn x []
     --> x
@@ -534,11 +540,11 @@ Destructuring using case expressions
     List.foldr 0 (+) list
     --> List.sum list
 
-    List.foldr 1 (*) list
-    --> List.product list
-
     List.foldr initial (+) list
     --> initial + List.sum list
+
+    List.foldr 1 (*) list
+    --> initial + List.product list
 
     List.foldr initial (*) list
     --> initial + List.product list
@@ -548,6 +554,12 @@ Destructuring using case expressions
 
     List.foldr False (&&) list
     --> False
+
+    List.foldr False (||) list
+    --> List.any identity list
+
+    List.foldr True (||) list
+    --> True
 
     List.foldr fn x []
     --> x
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -3868,39 +3868,36 @@ listFoldAnyDirectionChecks foldOperationName checkInfo =
                                        Maybe.andThen
                                         (\alwaysResult ->
                                             if isIdentity checkInfo.lookupTable alwaysResult then
-                                                Just ()
+                                                Just
+                                                    [ Rule.errorWithFix
+                                                        { message = "The call to List." ++ foldOperationName ++ " will result in the initial accumulator"
+                                                        , details = [ "You can replace this call by the initial accumulator." ]
+                                                        }
+                                                        checkInfo.fnRange
+                                                        (case checkInfo.thirdArg of
+                                                            Nothing ->
+                                                                [ Fix.replaceRangeBy
+                                                                    { start = checkInfo.parentRange.start
+                                                                    , end = (Node.range checkInfo.firstArg).end
+                                                                    }
+                                                                    "always"
+                                                                ]
+
+                                                            Just _ ->
+                                                                [ Fix.removeRange
+                                                                    { start = (Node.range initialArgument).end
+                                                                    , end = checkInfo.parentRange.end
+                                                                    }
+                                                                , Fix.removeRange
+                                                                    { start = checkInfo.parentRange.start
+                                                                    , end = (Node.range initialArgument).start
+                                                                    }
+                                                                ]
+                                                        )
+                                                    ]
 
                                             else
                                                 Nothing
-                                        )
-                                    |> Maybe.map
-                                        (\() ->
-                                            [ Rule.errorWithFix
-                                                { message = "The call to List." ++ foldOperationName ++ " will result in the initial accumulator"
-                                                , details = [ "You can replace this call by the initial accumulator." ]
-                                                }
-                                                checkInfo.fnRange
-                                                (case checkInfo.thirdArg of
-                                                    Nothing ->
-                                                        [ Fix.replaceRangeBy
-                                                            { start = checkInfo.parentRange.start
-                                                            , end = (Node.range checkInfo.firstArg).end
-                                                            }
-                                                            "always"
-                                                        ]
-
-                                                    Just _ ->
-                                                        [ Fix.removeRange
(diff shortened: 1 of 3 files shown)


src/Simplify.elm
```diff
@@ -3605,6 +3695,228 @@ listIndexedMapChecks { lookupTable, fnRange, firstArg } =
...
+
+
+listFoldrChecks : CheckInfo -> List (Error {})
+listFoldrChecks checkInfo =
+    listFoldAnyDirectionChecks "foldr" checkInfo
+
+
+listFoldAnyDirectionChecks : String -> CheckInfo -> List (Error {})
+listFoldAnyDirectionChecks foldOperationName checkInfo =
+    -- TODO discuss adding
```
lue-bird, line 3710: comments starting from here

src/Simplify.elm
```diff
@@ -3605,6 +3719,228 @@ listIndexedMapChecks { lookupTable, fnRange, firstArg } =
...
+        ( Just (Node initialArgumentRange _), Just (Node _ (Expression.ListExpr [])) ) ->
+            [ Rule.errorWithFix
+                { message = "The call to List." ++ foldOperationName ++ " will result in the initial accumulator"
+                , details = [ "You can replace this call by the initial accumulator." ]
+                }
+                checkInfo.fnRange
+                [ Fix.removeRange
+                    { start = initialArgumentRange.end, end = checkInfo.parentRange.end }
+                , Fix.removeRange
+                    { start = checkInfo.parentRange.start, end = initialArgumentRange.start }
```
jfmengels, line 3753: Can you change the order of these from earlier in the file to later in the file? This was confusing to me :sweat_smile: 

This will be sorted that way anyway, so no need to do it the opposite way for performance.

src/Simplify.elm
```diff
@@ -3605,6 +3719,228 @@ listIndexedMapChecks { lookupTable, fnRange, firstArg } =
...
+
+listFoldrChecks : CheckInfo -> List (Error {})
+listFoldrChecks checkInfo =
+    listFoldAnyDirectionChecks "foldr" checkInfo
+
+
+listFoldAnyDirectionChecks : String -> CheckInfo -> List (Error {})
+listFoldAnyDirectionChecks foldOperationName checkInfo =
+    -- TODO discuss adding
+    -- List.fold (&|) initial list --> initial &| (List.all identity list)
```
jfmengels, line 3735: Hmm... I'd say no just because of code style (especially with piping) :shrug:

src/Simplify.elm
```diff
@@ -3605,6 +3719,228 @@ listIndexedMapChecks { lookupTable, fnRange, firstArg } =
...
+    --         and piping versions
+    -- List.fold (&|) initial --> (List.all identity >> (&|) initial)
+    --     Are those simplifications? arguable, I'm leaning no
```
jfmengels, line 3738: No for this one. I'd say we never want to introduce `>>` or `<<`, because that is often a bit hard to read for beginners.

EDIT: You said the same thing below

src/Simplify.elm
```diff
@@ -3605,6 +3719,228 @@ listIndexedMapChecks { lookupTable, fnRange, firstArg } =
...
+                checkInfo.fnRange
+                [ Fix.removeRange
+                    { start = initialArgumentRange.end, end = checkInfo.parentRange.end }
+                , Fix.removeRange
+                    { start = checkInfo.parentRange.start, end = initialArgumentRange.start }
+                ]
+            ]
+
+        _ ->
+            checkInfo.secondArg
```
jfmengels, line 3758: You could simplify this function by pattern matching on the second and third argument separately. This way you don't need to do `Maybe.andThen` in the `_` case.

```elm
listFoldAnyDirectionChecks : String -> CheckInfo -> List (Error {})
listFoldAnyDirectionChecks foldOperationName checkInfo =
  case checkInfo.secondArg of
    Just initialArgument ->
      case checkInfo.thirdArg of
        Just (Node _ (Expression.ListExpr [])) ->
            -- error about it equating the initial value
        Nothing ->
            -- ...
    Nothing ->
      []
```

src/Simplify.elm
```diff
@@ -3605,6 +3719,228 @@ listIndexedMapChecks { lookupTable, fnRange, firstArg } =
...
+                        let
+                            numberBinaryOperationChecks : { identity : Int, two : String, list : String } -> List (Error {})
+                            numberBinaryOperationChecks operation =
+                                [ Rule.errorWithFix
+                                    { message = "Use List." ++ operation.list ++ " instead"
+                                    , details =
+                                        [ "Using List." ++ foldOperationName ++ " (" ++ operation.two ++ ") " ++ String.fromInt operation.identity ++ " is the same as using List." ++ operation.list ++ "." ]
+                                    }
+                                    checkInfo.fnRange
+                                    (if getUncomputedNumberValue initialArgument == Just (Basics.toFloat operation.identity) then
```
jfmengels, line 3775: This fix argument could definitely be extracted to a separate function

src/Simplify.elm
```diff
@@ -3605,6 +3719,228 @@ listIndexedMapChecks { lookupTable, fnRange, firstArg } =
...
+                                            , end = (Node.range initialArgument).end
+                                            }
+                                            ("List." ++ operation.list)
+                                        ]
+
+                                     else
+                                        case checkInfo.thirdArg of
+                                            Nothing ->
+                                                -- fold op initial --> (List.op >> op initial)
+                                                -- TODO Can we call this a simplification?
```
jfmengels, line 3787: I'd say no unfortunately :/

src/Simplify.elm
```diff
@@ -3605,6 +3719,228 @@ listIndexedMapChecks { lookupTable, fnRange, firstArg } =
...
+                                                    }
+                                                , Fix.insertAt (Node.range initialArgument).start
+                                                    (" >> (" ++ operation.two ++ ") ")
+                                                , Fix.insertAt checkInfo.parentRange.start
+                                                    ("(List." ++ operation.list)
+                                                ]
+
+                                            Just _ ->
+                                                if checkInfo.usingRightPizza then
+                                                    -- list |> fold op initial --> (list |> List.op) + initial
```
jfmengels, line 3806: Does this not create problems like if you have
```elm
a
  |> f
  |> List.foldl (+) initial
  |> g
-->
(a
  |> f
  |> List.sum
) + initial
  |> g
```
I'm always worried that this doesn't work or not. Like is this `(...) + (initial |> g)` or is it `((...) + initial) |> g`? When I write this code manually, I always add parentheses because it scares me that it's not going to do what I hoped.

jfmengels (changes requested):
I haven't had time to review all of it, I'll do the rest later :)
Thanks for working on this @lue-bird, I really appreciate it!

src/Simplify.elm
```diff
@@ -501,6 +501,72 @@ Destructuring using case expressions
...
+    List.foldl True (&&) list
+    --> List.all identity list
+
+    List.foldl False (&&) list
+    --> False
+
+    List.foldl False (||) list
+    --> List.any identity list
+
+    List.foldl True (||) list
```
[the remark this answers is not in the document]
lue-bird: Having a different order in all my packages ([example](https://dark.elm.dmy.fr/packages/lue-bird/elm-linear-direction/latest/List-Linear#foldFrom)) seems to have confused me ^^

src/Simplify.elm
lue-bird: 🤯 It's those little things that elm-review silently does

src/Simplify.elm
lue-bird: Yep that's better when we don't do fixes for initial = Nothing 👍

lue-bird:
Thanks for your feedback!
I'll move this last TODO from the code to here for discussion:
```elm
List.fold f x [ a ] --> f (a) x
```
> potentially quite ugly

The result would be the same for any pipes. I'd say it looks fine

src/Simplify.elm
lue-bird: do you mean something like
```elm
let fixWith fixes = Rule.errorWithFix ... fixes
in
```

src/Simplify.elm
lue-bird: Good catch

lue-bird:
Also, how about only showing simplifications for `foldl` in the rule documentation?

src/Simplify.elm
```diff
@@ -3860,7 +3866,18 @@ listFoldAnyDirectionChecks foldOperationName checkInfo =
                 numberBinaryOperationChecks { two = "+", list = "sum", identity = 0 }
 
             else if isBinaryOperation "*" checkInfo checkInfo.firstArg then
-                numberBinaryOperationChecks { two = "*", list = "product", identity = 1 }
+                if getUncomputedNumberValue initialArgument == Just 0 then
+                    [ Rule.errorWithFix
+                        { message = "The call to List." ++ foldOperationName ++ " (*) 0 will result in 0."
```
jfmengels, line 3871: Considering https://github.com/jfmengels/elm-review-simplify/issues/58, this is not really true, right?
```elm
List.foldl (*) 0 [0/0]
--> NaN
```

jfmengels:
> Also, how about only showing simplifications for foldl in the rule documentation?

Yeah. I think doing something like
```
-- The following simplifications for List.foldl also work for List.foldr
List.foldl (+) 0 list
--> List.sum list
```

src/Simplify.elm
lue-bird: Yes. Depending on what we decide on that issue, we'll have to do the same here

tests/SimplifyTest.elm
```diff
@@ -6959,6 +6961,1488 @@ a = False
...
+                            , under = "List.foldl"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = x
+"""
+                        ]
+        , test "should replace List.foldl (always identity) x by always x" <|
+            \() ->
+                """module A exposing (..)
+a = List.foldl (always identity) x
```
jfmengels, line 7057: Handling all of these variations is super nice :raised_hands:

tests/SimplifyTest.elm
```diff
@@ -6959,6 +6961,1488 @@ a = False
...
+                        [ Review.Test.error
+                            { message = "The call to List.foldl will result in the initial accumulator"
+                            , details = [ "You can replace this call by the initial accumulator." ]
+                            , under = "List.foldl"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = always x
+"""
+                        ]
+        , Test.concat listFoldlSumTests
```
jfmengels, line 7070: I didn't even know `Test.concat` existed! Though I'd say it's probably nicer to have a describe for each one (at their declaration, not here)

tests/SimplifyTest.elm
```diff
@@ -6959,6 +6961,1488 @@ a = False
...
+                        , under = "List.foldl"
+                        }
+                        |> Review.Test.whenFixed """module A exposing (..)
+a = False
+"""
+                    ]
+    , test "should replace List.foldl (&&) True by List.all identity" <|
+        \() ->
+            """module A exposing (..)
+a = List.foldl (&&) True
```
jfmengels, line 7253: I think we could use one test like this but where the list argument is specified, so we can be sure that one doesn't removed by accident. (Same for similar test suites)

jfmengels (changes requested):
I think we're almost there :heart:

tests/SimplifyTest.elm
lue-bird: Agreed. Wasn't sure since your tests for a single function were never split off into separate describes I think

jfmengels (approved):

jfmengels:
Alright! This looks good. Thanks for the hard work!

Maybe next time let's have smaller PRs (with more iteration / subsquent PRs), that will make it easier to review :grin: 

I'll try this out at work beginning of next week and release it if I don't see problems :+1:

PR #60, opened 2023-01-20
Add tests for String.length simplifications

lue-bird:
Couldn't find tests for `String.length` simplifications which I found spooky. Here are some.
[elm-syntax itself doesn't have many string parsing tests it seems](https://github.com/stil4m/elm-syntax/blob/master/tests/Elm/Parser/ExpressionTests.elm#L41-L60) (here [tokens](https://github.com/stil4m/elm-syntax/blob/master/tests/Elm/Parser/TokenTests.elm#L146-L173)); let's add other kinds of characters you can think of


jfmengels:
This is great! Thanks :relaxed:

PR #62, opened 2023-01-20
Add List.sum simplifications

lue-bird:
```elm
List.sum [] --> 0
List.sum [ a ] --> a
```
Simplifications for `product`, `minimum`, `maximum` will follow with exactly the same fixes

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -3725,11 +3725,11 @@ listSumChecks checkInfo =
     case Node.value checkInfo.firstArg of
         Expression.ListExpr [] ->
             [ Rule.errorWithFix
-                { message = "Using List.sum on [] will result in []"
-                , details = [ "You can replace this call by []." ]
+                { message = "Using List.sum on [] will result in 0"
+                , details = [ "You can replace this call by 0." ]
                 }
                 checkInfo.fnRange
-                [ Fix.replaceRangeBy checkInfo.parentRange "[]" ]
+                [ Fix.replaceRangeBy checkInfo.parentRange "0" ]
             ]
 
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -3713,6 +3720,35 @@ listIndexedMapChecks { lookupTable, fnRange, firstArg } =
                     []
 
 
+listSumChecks : CheckInfo -> List (Error {})
+listSumChecks checkInfo =
+    case Node.value checkInfo.firstArg of
+        Expression.ListExpr [] ->
+            [ Rule.errorWithFix
+                { message = "Using List.sum on [] will result in []"
```
jfmengels, line 3728: This should say 0

src/Simplify.elm
```diff
@@ -3713,6 +3720,35 @@ listIndexedMapChecks { lookupTable, fnRange, firstArg } =
...
+                , details = [ "You can replace this call by []." ]
+                }
+                checkInfo.fnRange
+                [ Fix.replaceRangeBy checkInfo.parentRange "[]" ]
```
jfmengels, line 3732: This should also be 0

lue-bird: absolutely right

lue-bird: To catch these "dumb mistakes", there could be checks on whether the simplified version doesn't have more narrow types for anything... Which is overkill and needs elm type inference to be complete

jfmengels: Yeah. In an ideal world, we would have the Elm compiler compiling the code. Maybe with `Elm-in-Elm` or something? :sweat_smile:

jfmengels (approved):
Nice :)

PR #66, opened 2023-01-21
Clean up simplification summary

lue-bird:
For the rule documentation simplification summary, the PR only keeps one of multiple with the same idea.

Examples:
- ```elm
  List.concatMap identity --> List.concat
  List.concatMap identity list --> List.concat list
  ```
  where we only show
  ```elm
  List.concatMap identity list --> List.concat list
  ```
- ```elm
  List.map identity --> identity
  List.map identity list --> list
  ```
  where we only show
  ```elm
  List.map identity list --> list
  ```
- ```elm
  List.filter (always False) --> always []
  List.filter (always False) list --> []
  ```
  where we only show
  ```elm
  List.filter (always False) list --> []
  ```
- ```elm
  negate >> negate --> identity
  negate (negate x) --> x
  ```
  where we only show
  ```elm
  negate (negate x) --> x
  ```

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -49,9 +49,6 @@ Below is the list of all kinds of simplifications this rule applies.
     not (not x)
     --> x
 
-    not >> not
-    --> identity
-
 
 ### Comparisons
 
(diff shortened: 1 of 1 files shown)


jfmengels (approved):
The list is getting pretty long yeah. Sounds good to me :+1:

PR #68, opened 2023-01-21
Add String.fromList simplifications

lue-bird:
```elm
String.fromList [] --> ""
String.fromList [ a ] --> String.fromChar a
```

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -245,6 +245,12 @@ Destructuring using case expressions
     "a" ++ ""
     --> "a"
 
+    String.fromList []
+    --> ""
+
+    String.fromList [ a ]
+    --> String.fromChar a
+
     String.isEmpty ""
     --> True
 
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -2922,6 +2929,33 @@ reportEmptyListFirstArgument ( ( moduleName, name ), function ) =
...
+        Expression.ListExpr [] ->
+            [ Rule.errorWithFix
+                { message = "Calling String.fromList [] will result in " ++ emptyStringAsString
+                , details = [ "You can replace this call by " ++ emptyStringAsString ++ "." ]
+                }
+                checkInfo.fnRange
+                [ Fix.replaceRangeBy checkInfo.parentRange emptyStringAsString ]
+            ]
+
+        Expression.ListExpr ((Node keep _) :: []) ->
```
jfmengels, line 2944: One thing that we could do later on, is the notice calls like `String.fromList (List.singleton x)` and consider those to be the same as `String.fromList [ x ]`. But that's for later (and for a lot more things)

jfmengels:
All looks good :+1:

src/Simplify.elm
lue-bird: absolutely.

PR #69, opened 2023-01-21
Add all arguments to CheckInfo

lue-bird:
Refactoring internals. Instead of checking for argument counts one to three, we collect all arguments.

What's the motivation?

```elm
listMapNChecks : { n : Int } -> CheckInfo -> List (Error {})
listMapNChecks { n } checkInfo =
    if ... any argument after the first is empty ... then ...
```
The new implementation also cleans up some `CheckInfo` creation code which is nice.

I tried not to change the check implementations in the process, like switching to `keepOnlyFix` where possible. Having all tests pass first try makes me confident this didn't change any behavior.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -689,6 +689,7 @@ All of these also apply for `Sub`.
 
 -}
 
+import Array exposing (Array)
 import Dict exposing (Dict)
 import Elm.Docs
 import Elm.Syntax.Expression as Expression exposing (Expression, RecordSetter)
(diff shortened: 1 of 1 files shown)


src/Simplify.elm
```diff
@@ -1575,13 +1538,22 @@ type alias CheckInfo =
     , inferredConstants : ( Infer.Inferred, List Infer.Inferred )
     , parentRange : Range
     , fnRange : Range
-    , firstArg : Node Expression
-    , secondArg : Maybe (Node Expression)
-    , thirdArg : Maybe (Node Expression)
     , usingRightPizza : Bool
+    , firstArg : Node Expression
+    , argsAfterFirst : Array (Node Expression)
```
lue-bird, line 1543: Not that it matters much- Should we represent the arguments as an `Array` or a `List`?

jfmengels:
I like the consistency of using `checkInfo` and not destructuring it in the arguments.
I like/don't mind the `toCheckInfo` function, it simplifies things a bit.

What I'm unsure about is the representation of args as an array (or list), rather than separate fields. I agree that the full list of arguments can be useful and we should therefore add it once we need it.
But I don't know if in the regular cases, where we care specifically about the second or third argument, whether this will have performance impacts. I'm pretty sure that creating an array and `Array.get`ting from it is slower than what we were doing before (storing as a field).
Since accessing the second and third args are such common cases, and accessing the other ones is so rare, I'm thinking we should store the full list of arguments in `CheckInfo` but keep the second and third arguments as they were before :thinking:

src/Simplify.elm
```diff
@@ -4913,8 +4902,13 @@ resultWithDefaultChecks checkInfo =
...
 
 collectionFilterChecks : Collection -> CheckInfo -> List (Error {})
-collectionFilterChecks collection ({ parentRange, fnRange, firstArg, secondArg } as checkInfo) =
-    case Maybe.andThen (collection.determineSize checkInfo.lookupTable) checkInfo.secondArg of
+collectionFilterChecks collection checkInfo =
+    let
+        collectionArg : Maybe (Node Expression)
+        collectionArg =
+            secondArg checkInfo
+    in
```
lue-bird, line 4910: I'm not fully convinced since you usually only access second, third once in the ...checks functions and not repeatedly.
I agree tho that converting to an `Array` is not worth the cost

jfmengels: Right. Let's try this out and see how we like it in practice

PR #70, opened 2023-01-22
Add List.map2-5 simplifications

lue-bird:
```elm
-- same for up to List.map5 when any list is empty
List.map2 fn xs [] --> []
List.map2 fn [] ys --> []
```
Having #69, implementing this was easy

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -4531,11 +4531,26 @@ listMapNChecks { n } checkInfo =
             , details = [ "You can replace this call by []." ]
             }
             checkInfo.fnRange
-            [ Fix.replaceRangeBy checkInfo.parentRange "[]" ]
+            [ Fix.replaceRangeBy checkInfo.parentRange
+                (multiAlways (n - List.length checkInfo.argsAfterFirst) "[]")
+            ]
         ]
 
     else
         []
+
+
+multiAlways : Int -> String -> String
+multiAlways alwaysCount alwaysResultExpressionAsString =
+    case alwaysCount of
+        0 ->
+            alwaysResultExpressionAsString
+
+        1 ->
+            "always " ++ alwaysResultExpressionAsString
+
+        alwaysCountPositive ->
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -4512,6 +4523,21 @@ listDropChecks checkInfo =
...
 
+listMapNChecks : { n : Int } -> CheckInfo -> List (Error {})
+listMapNChecks { n } checkInfo =
+    if List.any (\(Node _ expression) -> expression == Expression.ListExpr []) checkInfo.argsAfterFirst then
+        [ Rule.errorWithFix
+            { message = "Using List.map" ++ String.fromInt n ++ " with any list being [] will result in []"
+            , details = [ "You can replace this call by []." ]
+            }
+            checkInfo.fnRange
+            [ Fix.replaceRangeBy checkInfo.parentRange "[]" ]
```
jfmengels, line 4534: The fix is incorrect for when there are less arguments than necessary. For instance
```elm
List.map2 f []
```
currently gets fixed to
```elm
[]
```
but should be something like
```elm
(\_ -> [])
```
with one `_` per missing argument. Or similarly using `always`.

I think using `always []` would be a fine fix if there is one missing argument, but I'm unsure what would be "nice enough" when there are more missing parameters. I'm thinking having `always (always (always []))` would be pretty unreadable, so in that case `\_ _ _ -> []` would be nicer? Or we could avoid writing a fix?

Let's add a unit test for this as well.

lue-bird: Yes. Also good catch with the parens which are required if the `List.mapN` call is right-pizza-ed into another function

jfmengels: I'm not fond of `always` myself, but mostly because I would rarely write it myself.

jfmengels:
Very nice, especially the details change. We should probably improve that actually, since in a lot of places we say "you can replace this by x" when the fix replaces the code by "always x" or something.

PR #71, opened 2023-01-22
Add List.append simplifications

lue-bird:
```elm
List.append [] ys --> ys
List.append xs [] --> xs
List.append [ a, b ] [ c ] --> [ a, b, c ]
```


jfmengels (approved):
LGTM :+1:

Issue #2, opened 2021-04-22
Simplifications to add


jfmengels:
Here are some of the list of simplifications I'd like to support.

If you'd like to work on some of these, please reply in a comment! If you want to suggest new ones, please **open a new issue**.

```elm
### Language

case x of
  A -> a
  B -> a
-- a (unless if the cases reference a value extracted from the pattern, or if nothing is ever extracted I guess?)

case x of
  Anything _ -> a
--> a


### Boolean

not (not a)
-- a

a |> not |> not
-- a

### String

String.replace x x str
--> str

String.replace str1 str2 ""
--> ""

String.slice 3 3 str
--> ""

String.slice 0 n str
--> String.left n str


### Numbers

a |> negate |> negate
--> a

### List

List.foldl (+) 0 data
--> List.sum data

List.foldl (\a b -> a + b) 0 data
--> List.sum data

List.foldl (\a -> (+) a) 0 data
--> List.sum data

--???
List.foldl (+) 5 data
--> 5 + (List.sum data)

a |> List.map f |> List.filterMap identity
--> a |> List.filterMap f

List.map f >> List.filterMap identity
--> List.filterMap f

List.filterMap identity [ Just a, Just b ]
--> [ a, b ]

-- Consider lists made with List.singleton to have length 1
List.singleton a |> List.isEmpty
--> False

List.indexedMap (\(_, a) ... -> ...)
--> List.map (\(a) ... -> ...)

List.indexedMap Tuple.second x
--> x

List.indexedMap (Tuple.second >> fn)
-- List.map fn

List.member a [a, b]
-- True

List.partition fn []
-- ( [], [] )

### Html

Html.Attributes.classList [x, y, (z, False)]
-- Html.Attributes.classList [x, y]

Html.Attributes.classList [(x, True), (y, True)]
-- Html.Attributes.class (String.join " " [x, y])

### Dict

[] |> Dict.fromList
-- Dict.empty

-- Probably a lot basic ones

dict |> Dict.toList |> List.map Tuple.second
-- dict |> Dict.values

dict |> Dict.toList |> List.map (\(_, value) -> value)
-- dict |> Dict.values

dict |> Dict.toList |> List.map Tuple.first
-- dict |> Dict.keys

dict |> Dict.toList |> List.map (Tuple.second >> f)
-- dict |> Dict.values |> List.map (f)

### Set

[] |> Set.fromList
-- Set.empty

Set.map identity set
-- set

[...]

PR #75, opened 2023-01-23
Don't apply List.indexedMap's checks to List.intersperse

### LINKED ISSUE #72: List.intersperse only accidentally works
https://github.com/jfmengels/elm-review-simplify/blob/main/src/Simplify.elm#L1590
should be changed to
```elm
, reportEmptyListSecondArgument ( ( [ "List" ], "intersperse" ), listIntersperseChecks )
```
with
```elm
listIntersperseChecks : CheckInfo -> List (Error {})
listIntersperseChecks _ =
    []
```

jfmengels:
Fixes #72

I went for a lambda rather than a separate function, because I can predict that looking at that function would confuse me as to why it exists :sweat_smile:

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1594,7 +1594,7 @@ functionCallChecks =
         , reportEmptyListFirstArgument ( ( [ "List" ], "concat" ), listConcatChecks )
         , reportEmptyListSecondArgument ( ( [ "List" ], "concatMap" ), listConcatMapChecks )
         , reportEmptyListSecondArgument ( ( [ "List" ], "indexedMap" ), listIndexedMapChecks )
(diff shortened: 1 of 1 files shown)


lue-bird (approved):
fine. Maybe it makes sense to extract the whole `reportEmptyListSecondArgument ( ( [ "List" ], "intersperse" ), \_ -> [] )` as `listIntersperseChecks`?
(If you're already in the grind of review-simlpify, mind looking over https://github.com/jfmengels/elm-review-simplify/issues/2#issuecomment-1381538258 to the simplification suggestions)

jfmengels:
I think it's okay as it is. I'd say let's revisit it when we find recurring patterns similar to what other functions do, and change this then to what makes most sense.

I'll take a look at the other issue now.

PR #76, opened 2023-01-24
Add Set fold via toList simplifications

lue-bird:
```elm
-- same for foldr
List.foldl f x (Set.toList set)
--> Set.foldl f x set
```
Composition like `Set.foldl >> List.foldl f x` works as well.
I put this in the summary section `### Sets` but internally it's checked from `( [ "List" ], "fold..." )`

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1734,6 +1734,8 @@ compositionChecks =
     , resultMapCompositionChecks
     , filterAndMapCompositionCheck
     , concatAndMapCompositionCheck
+    , foldAndSetToListCompositionChecks "foldl"
+    , foldAndSetToListCompositionChecks "foldr"
     ]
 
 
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -4181,6 +4183,36 @@ listFoldAnyDirectionChecks foldOperationName checkInfo =
...
+foldAndSetToListCompositionChecks foldOperationName checkInfo =
+    let
+        ( earlier, later ) =
+            if checkInfo.fromLeftToRight then
+                ( checkInfo.left, checkInfo.right )
+
+            else
+                ( checkInfo.right, checkInfo.left )
+    in
+    case getSpecificFunctionCall ( [ "List" ], foldOperationName ) checkInfo.lookupTable later of
```
jfmengels, line 4196: I *think* we should be checking for the number of arguments here. Someone could write code like
```elm
List.foldr f << Set.toList -- instead of List.foldr f x << Set.toList
--> which would be transformed to
Set.foldr f
```
which wouldn't be correct I believe.

jfmengels (changes requested):
Sorry for the delay, this looks great!
There's one part where I'm worried that the fix is incorrect though :thinking:

jfmengels:
Awesome, this is a very nice change to get in :blush:

PR #77, opened 2023-01-30
fix #58 (* 0.0)

### LINKED ISSUE #58: `n * 0` isn't always `0`
`multiplyChecks`, `listFoldrChecks`, `listFoldlChecks` use `getUncomputedNumberValue` to check for `0`. Returning `Float * 0` however does not give the correct answer for certain very rare edge cases:
```elm
 1 / 0 --> Infinity
-1 / 0 --> -Infinity
 0 / 0 --> NaN
```
which multiplied with `0` all give `NaN`.

I think the rule helps you _a lot more often_ than it bites you, so if you think a few very rare wrong fixes are the price, just close the issue.

Possible steps to improve the situation:
- only report an error without fixes or nothing at all for `Expression.Floatable 0` (This is not what `Simplify` usually does)
- keep the current behavior but try to have separate checks on NaN and Infinity (like if NaN is multiplied by something, the result is NaN)
- (suggested by @gampleman) keep the current behavior but warn about NaN in the message and suggest replacing it with an explicit check for NaN if that behaviour was the intended one

lue-bird:
Multiplication by 0.0 now doesn't get any fixes as suggested in #58 and reports enough details to know how to proceed.
Let's tweak the error message and info further 🎈.

One thing I'm uncertain about: `0` always gets parsed as `Integer` (not `Floatable`) _but_ it satisfies `number`. `number` could be used as `Float` tho, which could again produce NaN, (-)Infinity but which would be fixed to `0` by elm-review-simplify.
My reaction would be to never fix to `0` unless we have type inference activated. WDYT?

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1887,8 +1887,9 @@ multiplyChecks { leftRange, rightRange, left, right } =
                                     [ "Multiplying by 0 will turn finite numbers into 0 and keep NaN and (-)Infinity"
                                     , "Most likely, multiplying by 0 was unintentional and you had a different factor in mind."
                                     , """If you do want the described behavior, though, make your intention clear for the reader
-by explicitly checking for [`Basics.isNaN`](https://package.elm-lang.org/packages/elm/core/latest/Basics#isNaN)
-and [`Basics.isInfinite`](https://package.elm-lang.org/packages/elm/core/latest/Basics#isInfinite)"""
+by explicitly checking for `Basics.isNaN` and `Basics.isInfinite`."""
+                                    , """Basics.isNaN: https://package.elm-lang.org/packages/elm/core/latest/Basics#isNaN
+Basics.isInfinite: https://package.elm-lang.org/packages/elm/core/latest/Basics#isInfinite"""
                                     ]
                                 }
                                 (getRange ())
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1870,33 +1870,43 @@ multiplyChecks : OperatorCheckInfo -> List (Error {})
 multiplyChecks { parentRange, leftRange, rightRange, left, right } =
     findMap
         (\( node, getRange ) ->
-            case getUncomputedNumberValue node of
-                Just value ->
-                    if value == 1 then
-                        Just
-                            [ Rule.errorWithFix
-                                { message = "Unnecessary multiplication by 1"
-                                , details = [ "Multiplying by 1 does not change the value of the number." ]
-                                }
-                                (getRange ())
-                                [ Fix.removeRange (getRange ()) ]
+            if getUncomputedNumberValue node == Just 1 then
+                Just
+                    [ Rule.errorWithFix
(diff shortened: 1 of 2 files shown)


jfmengels:
Why should we make a difference between `0` and `0.0`? We can have `NaN` as an integer, don't we? :thinking:

lue-bird:
> We can have NaN as an integer, don't we?

Yes, but I'd say that's probably more a bug than anything (`Int` math never returns `NaN`)

src/Simplify.elm
```diff
@@ -4152,12 +4154,17 @@ listFoldAnyDirectionChecks foldOperationName checkInfo =
 
                     else if isBinaryOperation "*" checkInfo checkInfo.firstArg then
                         if initialNumber == Just 0 then
-                            [ Rule.errorWithFix
-                                { message = "The call to List." ++ foldOperationName ++ " (*) 0 will result in 0."
-                                , details = [ "You can replace this call by 0." ]
+                            [ Rule.error
```
jfmengels, line 4157: I was thinking that the error message here didn't fit as nicely as in the other places.

But I'm thinking we maybe should not check for `initialNumber == Just 0` anymore in this situation. That way, we'll end up with `initialNumber * List.product list`. If `initialNumber` happens to be 0, then we'll end up with `0 * List.product list`, which will trigger the error in other places.
What do you think?

src/Simplify.elm
```diff
@@ -1882,20 +1879,25 @@ multiplyChecks { parentRange, leftRange, rightRange, left, right } =
...
-                                { message = "Multiplying by 0 equals 0"
-                                , details = [ "You can replace this value by 0." ]
+                            [ Rule.error
+                                { message = "Multiplication by 0 should be replaced"
+                                , details =
+                                    [ "Multiplying by 0 will turn finite numbers into 0 and keep NaN and (-)Infinity"
+                                    , "Most likely, multiplying by 0 was unintentional and you had a different factor in mind."
+                                    , """If you do want the described behavior, though, make your intention clear for the reader
+by explicitly checking for [`Basics.isNaN`](https://package.elm-lang.org/packages/elm/core/latest/Basics#isNaN)
+and [`Basics.isInfinite`](https://package.elm-lang.org/packages/elm/core/latest/Basics#isInfinite)"""
```
jfmengels, line 1891: There isn't any Markdown rendering in the details (that I can recall at least!), so I'm thinking it will render better like this:

```suggestion
                                    , """If you do want the described behavior, though, make your intention clear for the reader
by explicitly checking for `Basics.isNaN` and `Basics.isInfinite`."""
                                    , """Basics.isNaN: https://package.elm-lang.org/packages/elm/core/latest/Basics#isNaN
Basics.isInfinite: https://package.elm-lang.org/packages/elm/core/latest/Basics#isInfinite
```

(I'm open to suggestions on how to support Markdown/links/formatting in error details, but that's for another time :D )

jfmengels (approved):

jfmengels:
🙌

PR #78, opened 2023-01-31
Add List.head simplifications

lue-bird:
```elm
List.head [] --> Nothing
List.head (List.singleton a) --> Just a
List.head [ a, b, c ] --> Just a
List.head (a :: bToZ) --> Just a
```
(only the first and last simplifications are in the summary)

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -3916,22 +3916,21 @@ listHeadChecks checkInfo =
                 [ Fix.replaceRangeBy checkInfo.parentRange "Nothing" ]
             ]
 
-        Expression.ListExpr (head :: _) ->
-            let
-                headRange : Range
-                headRange =
-                    Node.range head
-            in
-            [ Rule.errorWithFix
-                listHeadExistsError
-                checkInfo.fnRange
-                (keepOnlyFix { parentRange = Node.range listArg, keep = headRange }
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -3875,6 +3882,65 @@ listAppendChecks checkInfo =
...
+listHeadChecks checkInfo =
+    let
+        justFirstElementError : Range -> List (Error {})
+        justFirstElementError keep =
+            [ Rule.errorWithFix
+                { message = "Using List.head on a list with a first element will result in Just that element"
+                , details = [ "You can replace this call by Just the first list element." ]
+                }
+                checkInfo.fnRange
+                (keepOnlyFix { parentRange = Node.range listArg, keep = keep }
```
jfmengels, line 3895: I'm thinking we may need to add parens around `keep`. There is `needsParens` to help determine that.
```elm
List.head [ f a ]
--> Just f a ❌
--> Just (f a) ✅
```

jfmengels (changes requested):

src/Simplify.elm
lue-bird: `needsParens` is :ok_hand:

jfmengels (approved):
Very nice :ok_hand:

PR #79, opened 2023-01-31
Add List.tail simplifications

lue-bird:
```elm
List.tail [] --> Nothing
List.tail (List.singleton a) --> Just []
List.tail [ a, b, c ] --> Just [ b, c ]
List.tail (a :: bToZ) --> Just bToZ
```
(only the first and last are shown in the summary)

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -2647,6 +2647,21 @@ getNotFunction lookupTable baseNode =
     getSpecificFunction ( [ "Basics" ], "not" ) lookupTable baseNode
 
 
+getListSingletonCall : ModuleNameLookupTable -> Node Expression -> Maybe { element : Node Expression }
+getListSingletonCall lookupTable expressionNode =
+    case getSpecificFunctionCall ( [ "List" ], "singleton" ) lookupTable expressionNode of
+        Just singletonCall ->
+            case singletonCall.argsAfterFirst of
+                [] ->
+                    Just { element = singletonCall.firstArg }
+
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -3944,19 +3966,95 @@ listHeadChecks checkInfo =
...
+                    [ Rule.errorWithFix
+                        listTailExistsError
+                        checkInfo.fnRange
+                        (keepOnlyFix { parentRange = listArgRange, keep = tailRange }
+                            ++ [ Fix.insertAt tailFirstRange.start "[ "
+                               , Fix.insertAt tailLastRange.end " ]"
+                               , Fix.replaceRangeBy checkInfo.fnRange "Just"
+                               ]
+                        )
+                    ]
```
jfmengels, line 4034: I think a simpler fix would be to remove the first element. So if you have `[ a, b, c ]`, remove `a, `.
```suggestion
                Node headRange _ :: (Node tailFirstRange _) :: _ ->
                    [ Rule.errorWithFix
                        listTailExistsError
                        checkInfo.fnRange
                        [ Fix.removeRange { start = headRange.start, end = tailFirstRange.start }
                        , Fix.replaceRangeBy checkInfo.fnRange "Just"
                        ]
                    ]
```
(I did not test the suggestion)

lue-bird: That's so clean

jfmengels:
One more off the list :heart:

PR #80, opened 2023-01-31
Add List.member simplifications

lue-bird:
```elm
List.member a [] --> False
List.member a (List.singleton a) --> True
List.member c (a :: b :: c :: dToZ) --> True
List.member a [ a, b, c ] --> True
List.member a [ b ] --> a == b
List.member a (List.singleton b) --> a == b
```
Only the first and last 3 are shown in the summary.
Apparently `member a [] --> False` was already implemented (https://github.com/jfmengels/elm-review-simplify/issues/2 suggested otherwise)

The possibility of the needle being NaN is silently ignored. The most accurate fix would be
```elm
List.member a [ a, b, c ] --> a == a
```
This merge request doesn't do that because above would be simplified to `True` anyway (at least currently)

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -4061,10 +4061,6 @@ listMemberChecks checkInfo =
         needleArg =
             checkInfo.firstArg
 
-        needleRange : Range
-        needleRange =
-            Node.range needleArg
-
         needleArgNormalized : Node Expression
         needleArgNormalized =
             Normalize.normalize checkInfo needleArg
@@ -4079,6 +4075,10 @@ listMemberChecks checkInfo =
     case secondArg checkInfo of
         Just listArg ->
             let
+                needleRange : Range
+                needleRange =
+                    Node.range needleArg
+
                 listMemberExistsError : List (Error {})
                 listMemberExistsError =
                     [ Rule.errorWithFix
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -428,6 +428,12 @@ Destructuring using case expressions
     List.tail (a :: bToZ)
     --> Just bToZ
 
+    List.member a []
+    --> False
+
+    List.member a [ a, b, c ]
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -4045,6 +4051,83 @@ listTailChecks checkInfo =
...
+listMemberChecks checkInfo =
+    let
+        needleArg : Node Expression
+        needleArg =
+            checkInfo.firstArg
+
+        isNeedle : Node Expression -> Bool
+        isNeedle element =
+            Normalize.compare checkInfo element needleArg
+                == Normalize.ConfirmedEquality
```
jfmengels, line 4064: We should probably try to use `Normalize.compareWithoutNormalization`, in order not to normalize `needleArg` once for each element in the list. Instead we should try to do it only once (but also only when we need to).

Something like this could work I think:

```elm
isNeedle needle =
  let normalizedNeedle = Normalize.normalize needle
  in
  \element -> Normalize.compareWithoutNormalization (Normalize.normalize element) normalizedNeedle
```

jfmengels (commented):
One suggestion, otherwise it looks good :+1: 

That said. Just like #58, should we even attempt to replace `List.member a [ a ]`? What if `a` is `NaN`? :/

lue-bird:
> That said. Just like https://github.com/jfmengels/elm-review-simplify/issues/58, should we even attempt to replace List.member a [ a ]? What if a is NaN? :/

I'd love to see member fixes as suggested above, like `member a [ a, b ] --> a == a`!
Since `a == a` is currently fixed to `True` (and `a /= a` to `False`) this currently doesn't make sense. If we open another merge request which only reports, doesn't fix `a == a` and `a /= a` we could also change `member`.

jfmengels (approved):
LGTM :+1: 
(Sorry, almost forgot about this PR :s )

PR #82, opened 2023-02-04
Add Html.Attributes.classList simplifications

lue-bird:
```elm
Html.Attributes.classList [ x, y, ( z, False ) ]
--> Html.Attributes.classList [ x, y ]

Html.Attributes.classList (x :: ( y, False ) :: zs)
--> Html.Attributes.classList (x :: zs)

Html.Attributes.classList (List.singleton ( y, False ))
--> Html.Attributes.classList []

Html.Attributes.classList (List.singleton ( onlyOneThing, True ))
--> Html.Attributes.class onlyOneThing

Html.Attributes.classList [ ( onlyOneThing, True ) ]
--> Html.Attributes.class onlyOneThing
```
Only the first and last simplifications are shown in the summary.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -717,6 +717,15 @@ All of these also apply for `Sub`.
     --> a
 
 
+### Html.Attributes
+
+    Html.Attributes.classList [ x, y, ( z, False ) ]
+    --> Html.Attributes.classList [ x, y ]
+
+    Html.Attributes.classList [ ( onlyOneThing, True ) ]
+    --> Html.Attributes.class onlyOneThing
+
+
 ### Parser
 
     Parser.oneOf [ a ]
@@ -1680,6 +1689,7 @@ functionCallChecks =
         , ( ( [ "Platform", "Sub" ], "batch" ), subAndCmdBatchChecks "Sub" )
         , ( ( [ "Platform", "Sub" ], "map" ), collectionMapChecks subCollection )
         , ( ( [ "Json", "Decode" ], "oneOf" ), oneOfChecks )
+        , ( ( [ "Html", "Attributes" ], "classList" ), htmlAttributesClassListChecks )
         , ( ( [ "Parser" ], "oneOf" ), oneOfChecks )
         , ( ( [ "Parser", "Advanced" ], "oneOf" ), oneOfChecks )
         ]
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -5127,6 +5137,204 @@ subAndCmdBatchChecks moduleName { lookupTable, parentRange, fnRange, firstArg }
...
+                    Nothing
+
+        singleElementListChecks : { a | element : Node Expression } -> List (Error {})
+        singleElementListChecks single =
+            case getTupleWithSecond True single.element of
+                Just tuple ->
+                    singleTrueChecks tuple
+
+                Nothing ->
+                    case getTupleWithSecond False single.element of
```
jfmengels, line 5162: If `getTupleWithSecond`/`getBool` returned a `Maybe Bool` with the boolean value, we could remove the number of similar computations and simplify the branching.

```elm
case getBoolInSecond single.element of
  Just True -> singleTrueChecks tuple
  Just False -> [ Rule.errorWithFix ... ]
  Nothing -> []
```

src/Simplify.elm
```diff
@@ -5127,6 +5137,204 @@ subAndCmdBatchChecks moduleName { lookupTable, parentRange, fnRange, firstArg }
...
+                            [ Rule.errorWithFix htmlAttributesClassListFalseElementError
+                                checkInfo.fnRange
+                                [ Fix.replaceRangeBy (Node.range listArg) listCollection.emptyAsString ]
+                            ]
+
+                        Nothing ->
+                            []
+
+        singleTrueChecks : { a | first : Node Expression } -> List (Error {})
+        singleTrueChecks { first } =
```
jfmengels, line 5173: I'm quite sad we can only do this check if it's there is only a single element in a lot of cases (because we don't have a function that takes a list of classes without conditions). But I'm thinking that if there are only literal strings in the first value, then we can simplify the expression to use `Html.Attributes.class`:

```elm
Html.Attributes.classList
  [ ( "some-class", True )
  , ( """other-class""", True )
  , ( "third-class", True )
  ]
--> Html.Attributes.class "some-class other-class third-class"
```
I think this is actually more common in old codebases, because I have seen classList used when only a single item was actually conditional.

This can be done in a separate PR. Actually, probably better done as a separate PR :smile:

src/Simplify.elm
```diff
@@ -5127,6 +5137,204 @@ subAndCmdBatchChecks moduleName { lookupTable, parentRange, fnRange, firstArg }
...
+                       ]
+                )
+            ]
+    in
+    case getListLiteral listArg of
+        Just (single :: []) ->
+            singleElementListChecks { element = single }
+
+        Just nonSingletonList ->
+            case findMapSurrounding (getTupleWithSecond False) nonSingletonList of
```
jfmengels, line 5190: The `tryMap` argument to `findMapSurrounding` is always `getTupleWithSecond`. You can probably hardcode that function in the implementation? Unless you think you'll use this for other simplifications?

Also, I'm trying to find a better name for this function but haven't figured one out yet (and depends on whether you agree on hardcoding `getTupleWithSecond` in the function)

jfmengels (changes requested):

src/Simplify.elm
lue-bird: > you think you'll use this for other simplifications?

Think for example removing `Nothing`s in `filterMap`

jfmengels: Alright, yeah that makes sense as a use-case 👍

src/Simplify.elm
```diff
@@ -5127,6 +5137,204 @@ subAndCmdBatchChecks moduleName { lookupTable, parentRange, fnRange, firstArg }
...
+
+        singleTrueChecks : { a | first : Node Expression } -> List (Error {})
+        singleTrueChecks { first } =
+            [ Rule.errorWithFix
+                { message = "Html.Attributes.classList with a single tuple paired with True can be replaced with Html.Attributes.class"
+                , details = [ "You can replace this call by Html.Attributes.class with the String from the single tuple list element." ]
+                }
+                checkInfo.fnRange
+                (keepOnlyFix { parentRange = Node.range listArg, keep = Node.range first }
+                    ++ [ Fix.replaceRangeBy checkInfo.fnRange "Html.Attributes.class"
```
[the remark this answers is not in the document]
lue-bird: Yes would be fantastic if there was some `ImportLookup = Dict ( ModuleName, String ) ModuleName` in the `CheckInfo` and a helper `toQualifiedString : ImportLookup -> ModuleName -> String -> String`.

I actually wrote this exact thing some time ago:
- https://github.com/lue-bird/generate-elm/blob/master/src/Elm/Generator.elm#L803
- https://github.com/lue-bird/generate-elm/blob/master/src/Review/Generate/Internal.elm#L178-L188
- https://github.com/lue-bird/generate-elm/blob/master/src/Review/Generate/Internal.elm#L1552-L1583

src/Simplify.elm
lue-bird: ok next PR

src/Simplify.elm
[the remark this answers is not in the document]
jfmengels: Yeah, I agree that this information could be quite useful for other rules. I probably rewrote this logic in some other rules as well.

I'm open to adding this functionality to the core package, but this could probably use some design, with some trying it out in several rules and seeing whether the information we care about is made available

For instance, in your example for `toQualifiedString`, the return type doesn't tell you whether you need to add an import of the module is not imported (useful for https://package.elm-lang.org/packages/dillonkearns/elm-review-html-to-elm/latest/ for instance). I think there was an issue somewhere as well to know where to inject a module import, which could be useful (though that's much easier to compute now that we have the full AST easily accessible now)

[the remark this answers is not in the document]
jfmengels: I'm still pondering whether we should probably fix this replacement module name issue before we merge this (or postpone this specific fix), because this will not work quite often. I know that this will break for my work codebase, whereas `Set` and `Dict` often will be imported without aliases.

lue-bird:
> fix this replacement module name issue before we merge this

How about we go the patch fix implementation route I suggested for this PR and overhaul all fixes in the import lookup PR? (I'm not a git wizard; leaving this branch open and catching up later seems oof)

> whether you need to add an import

I don't actually think elm-review-siplifiy should import stuff to fix things at all but I guess it could be useful _somewhere_, in which case your suggestion to return whether an import is needed and store a location to place imports at in the `CheckInfo` seems good.

IMO this can be easily introduced _once needed_.

PR #83, opened 2023-02-06
Don't simplify `String.slice 0` to `left`

lue-bird:
Removes
```elm
String.slice 0 n str --> String.left n str
```
because negative indices are accepted as end indices by `slice` but return `""` with `left`, so for example
```elm
"abcde" |> String.slice 0 -3 --> "ab"
"abcde" |> String.left -3 --> ""
```
> Reported by Décio Ferreira on slack

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -299,9 +299,6 @@ Destructuring using case expressions
     String.slice n n str
     --> ""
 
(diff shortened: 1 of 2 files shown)


jfmengels (approved):
:ok_hand:

PR #84, opened 2023-02-06
Add ImportLookup

lue-bird:
Implementing a lookup that helps figure out the module-local qualification for things https://github.com/jfmengels/elm-review-simplify/pull/82#discussion_r1096590841

- [x] collect `module` imports
- [x] collect enough info to know whether a variant is exposed with `Type(..)`
- [x] CheckInfo etc.
- [x] Collection etc.
- [x] correct existing fixes. There's _a lot_ of code replacing to a value or function, so correcting was a bit of a pain but it's done so it's done. (complete including `Just`, `Nothing`, `Err`, `Ok`, `True`, `False`). If you find other ones, let me know!
- [x] cover implicit default imports
- [x] merge imports

Overall, this is great because you can be sure that fixes (like those of #82) will always produce valid code. (well except when a module has a member with the same name as an implicitly imported thing)


lue-bird:
I hope existing code doesn't break since we now use `Rule.withContextFromImportedModules`.

src/Simplify/AstHelpers.elm
```diff
@@ -23,3 +25,16 @@ removeParensFromPattern node =
 
         _ ->
             node
+
+
+moduleHeaderExposing : Elm.Syntax.Module.Module -> Node Exposing
```
lue-bird, line 30: What's the idea behind the `AstHelper` module like only containing AST transformations? If not I'd probably want move most of the matchers and parsers from `Simplify` there.

jfmengels: To answer your question, I guess we could split the helpers a bit more yes. That would probably help make them more findable in practice?

But for this function, there is the `Elm.Module.Exposing.exposingList` which already does the same thing (well, it doesn't return a `Node`, but you seem to not care about that anyway)

src/Simplify.elm
```diff
@@ -953,24 +1036,129 @@ errorForUnknownIgnoredConstructor list =
...
+                                (Maybe.map
+                                    (\_ ->
+                                        variantNames
+                                    )
+                                )
+            }
+
+        _ ->
+            -- no change
+            contextReset
```
jfmengels, line 1159: Actually, we only need this reset only if the declaration is a `FunctionDeclaration`. So we could only change the context in that variant, and use the original context for the type declaration and for the wildcard case.

src/Simplify.elm
```diff
@@ -1041,9 +1229,42 @@ onlyErrors errors =
...
+                    (context.exposedVariants |> Dict.foldl (\_ -> Set.union) Set.empty)
+            }
+
+
 expressionVisitorHelp : Node Expression -> ModuleContext -> { errors : List (Error {}), rangesToIgnore : List Range, rightSidesOfPlusPlus : List Range, inferredConstants : List ( Range, Infer.Inferred ) }
 expressionVisitorHelp node context =
     let
+        importLookup : ImportLookup
+        importLookup =
+            moduleContextToImportLookup context
```
jfmengels, line 1266: This looks like a terribly expensive function to compute for every expression that we encounter.

I think it would be better to compute the `ImportLookup` once after we have visited all the imports. There is no visitor for that exact moment, so what we could do is compute it all in `fromProjectToModule` using the information `withFullAst`. We can then store an `ImportLookup` directly in the `ModuleContext` (I imagine, haven't fully checked).

src/Simplify.elm
```diff
@@ -2099,13 +2429,15 @@ negateNegateCompositionErrorMessage =
...
-negateCompositionCheck { lookupTable, fromLeftToRight, parentRange, left, right, leftRange, rightRange } =
+negateCompositionCheck { lookupTable, fromLeftToRight, parentRange, left, right, leftRange, rightRange, importLookup } =
     case Maybe.map2 Tuple.pair (getNegateFunction lookupTable left) (getNegateFunction lookupTable right) of
         Just _ ->
             [ Rule.errorWithFix
                 negateNegateCompositionErrorMessage
                 parentRange
-                [ Fix.replaceRangeBy parentRange "identity" ]
+                [ Fix.replaceRangeBy parentRange
+                    (qualifiedToString (qualify ( [ "Basics" ], "identity" ) importLookup))
```
jfmengels, line 2439: Is this ever not going to give us an unqualified `"identity"`? :thinking: 
I think for all the variables and types exposed through the Elm prelude, there is no need to try and do an import lookup, right?

jfmengels:
> I hope existing code doesn't break since we now use Rule.withContextFromImportedModules (I don't think it changes the way the rule is run here but you know best).

It won't in this case. That can only change the `ProjectContext` given to `fromProjectToModule`, where the previous context's fields were ignored anyway.

jfmengels (commented):
This look very good overall, great work!

Thanks for working on this (and sorry for the late reply, I somehow only noticed this last week when I was deep in talk-preparation).

I added a few comments, but they can be changed in separate PRs. The only one I feel is somewhat "urgent" is the computation of the import lookup in every file. That one would be nice to improve.

Shall I merge the PR as it is, or do you want to fix that one first?

tests/SimplifyTest.elm
```diff
@@ -164,6 +165,156 @@ a = 1
...
+                    |> Review.Test.run (rule defaults)
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "To fold a set, you don't need to convert to a List"
+                            , details = [ "Using Set.foldl directly is meant for this exact purpose and will also be faster." ]
+                            , under = "List.foldl"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+import Set exposing (foldl)
+a = foldl f x
```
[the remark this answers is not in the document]
lue-bird: It's not covered already and would be nice to add!

src/Simplify.elm
```diff
@@ -883,19 +937,29 @@ fromProjectToModule =
...
         |> Rule.withMetadata
         |> Rule.withSourceCodeExtractor
 
 
+foldProjectContexts : ProjectContext -> ProjectContext -> ProjectContext
+foldProjectContexts =
+    \a b ->
+        { initialContext
+            | exposedVariants = Dict.union a.exposedVariants b.exposedVariants
+        }
```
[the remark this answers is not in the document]
lue-bird: great idea

src/Simplify.elm
```diff
@@ -1400,6 +1628,105 @@ expressionVisitorHelp node context =
...
+    ]
+        |> Dict.fromList
+
+
+{-| Merge a given new import with an existing import lookup.
+This is strongly preferred over Dict.insert since the implicit default imports can be overridden
+-}
+insertImport : ModuleName -> { alias : Maybe ModuleName, exposed : Exposed } -> ImportLookup -> ImportLookup
+insertImport moduleName importInfoToAdd =
+    Dict.update moduleName
```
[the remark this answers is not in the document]
lue-bird: Did you find there other places except `insertImport`?

src/Simplify.elm
```diff
@@ -953,24 +1036,129 @@ errorForUnknownIgnoredConstructor list =
...
+                                variantNames
+
+                    else
+                        context.exposedVariants
+                            |> Dict.update variantTypeName
+                                (Maybe.map
+                                    (\_ ->
+                                        variantNames
+                                    )
+                                )
```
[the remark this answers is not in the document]
lue-bird: Why blow up the amount of variant names stored for types we don't expose (I think I'm misunderstanding the question)

src/Simplify.elm
lue-bird: ok

src/Simplify.elm
lue-bird: There might be a `module`-local shadowed function called `identity`

src/Simplify.elm
lue-bird: Thought about this as well – how to compute this once we finished going through all declarations... So why not do it at the end of [`withDeclarationListVisitor`](https://dark.elm.dmy.fr/packages/jfmengels/elm-review/latest/Review-Rule#withDeclarationListVisitor)?

lue-bird: >Shall I merge the PR as it is, or do you want to fix that one first?

Lemme fix that first

src/Simplify.elm
```diff
@@ -921,21 +985,40 @@ dependenciesVisitor typeNamesAsStrings dict _ =
...
+                |> Dict.fromList
     in
     ( if List.isEmpty unknownTypesToIgnore then
         []
 
       else
         [ errorForUnknownIgnoredConstructor unknownTypesToIgnore ]
-    , { customTypesToReportInCases = customTypesToReportInCases }
+    , { customTypesToReportInCases = customTypesToReportInCases
+      , exposedVariants = Dict.union context.exposedVariants dependencyExposedVariants
```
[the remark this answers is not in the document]
lue-bird: Why make assumptions that might bite you later (maybe implicit import module exposed variants will be in there on init in the future, who knows)

I'm not feeling strongly about this, so if you insist, I'll change it

tests/SimplifyTest.elm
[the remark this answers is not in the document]
jfmengels: Let's do it in another PR yes.

For (1), I think that maybe we can peek ahead using `withFullAst` once again :sweat_smile: 
(2 and 3) Ugh, yes, there's that as well... I think I might make an issue in the Elm compiler to forbid shadowing imports because they're pretty annoying? :sweat_smile:

src/Simplify.elm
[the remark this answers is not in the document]
jfmengels: That's fair, let's keep it as is.

src/Simplify.elm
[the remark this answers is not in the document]
jfmengels: I think I've seen it in other places, in other PRs IIRC? We can find them again later.

`insertImport` as you wrote it there works, but it's only better if you call it with 2 arguments. In practice you do call it with 3 arguments (`c |> insertImport a b` is turned into a simple function call `insertImport a b c` so that counts as 3).

src/Simplify.elm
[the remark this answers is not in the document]
jfmengels: Using `Dict.insert` always would not increase the number of variant names, since it will override the previous value. Also, I don't think it's in practice ever going to encounter the same `variantTypeName` twice. But maybe I'm in turn misunderstanding what this function does.

src/Simplify.elm
jfmengels: > So why not do it at the end of [withDeclarationListVisitor](https://dark.elm.dmy.fr/packages/jfmengels/elm-review/latest/Review-Rule#withDeclarationListVisitor)?

Yes, that works as well :+1:

src/Simplify.elm
jfmengels: > There might be a shadowed function called identity (as you have suggested; yet to be implemented)

Ugh. You're right... Shucks.

src/Simplify.elm
```diff
@@ -911,12 +911,12 @@ fromModuleToProject : Rule.ContextCreator ModuleContext ProjectContext
 fromModuleToProject =
     Rule.initContextCreator
         (\moduleContext ->
-            { initialContext
-                | exposedVariants =
-                    Dict.singleton moduleContext.moduleName
-                        (moduleContext.exposedVariants
-                            |> Dict.foldl (\_ -> Set.union) Set.empty
-                        )
+            { customTypesToReportInCases = initialContext.customTypesToReportInCases
```
jfmengels, line 913: This should be `Set.empty` to reduce the memory footprint (especially now with the file-system cache).

In practice, I never reference the initial context except in the rule's definition, to avoid it changing the behavior in other places.

jfmengels:
I'll merge this now, we can tackle the other comments later. Great work @lue-bird :heart:

src/Simplify.elm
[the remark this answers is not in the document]
lue-bird: It adds variant names for a specific type name – only if we need this information.
If we always insert the type and its variant names, even for type names that weren't added already to show they are exposed (this happens on ModuleContext init), we will collect and store information that's useless but storage space polluting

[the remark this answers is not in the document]
jfmengels: Ah gotcha, now I see it. Alright :+1:

PR #87, opened 2023-02-21
Fix #86

### LINKED ISSUE #86: List.foldl to List.sum fix loses function input
Running `npx elm-review --template jfmengels/elm-review-simplify/example` with the following module:
```elm
module Example exposing (oneOrLessTrue)

oneOrLessTrue : List Bool -> Bool
oneOrLessTrue bools =
    (bools |> List.map boolToInt |> List.foldl (+) 0) >= 1

boolToInt : Bool -> Int
boolToInt bool =
    if bool then
        1
    else
        0
```

produces the following fix:

```diff
  -- ELM-REVIEW ERROR --------------------------------------- src/Example.elm:5:37
  
  Simplify: Use List.sum instead
  
  4| oneOrLessTrue bools =
  5|     (bools |> List.map boolToInt |> List.foldl (+) 0) >= 1
                                         ^^^^^^^^^^
  
  Using List.foldl (+) 0 is the same as using List.sum.
  
  I think I can fix this. Here is my proposal:
  
  4| oneOrLessTrue bools =
- 5|     (bools |> List.map boolToInt |> List.foldl (+) 0) >= 1
+ 6|     (List.sum) >= 1
  6|
  
  √ Do you wish to apply this fix? ... no
```

It looks like the input to `List.foldl (+) 0` is getting lost, instead of being fed into `List.sum`.

lue-bird:
Corrects an error with pipelines and `List.fold` call simplifications: https://github.com/jfmengels/elm-review-simplify/issues/86
(tests included)

The fact that many issues are reported now is a sign at the quality of the rules before I added some I guess :balloon: 
Might be time for a PR adding some more tests and refactoring the old new rules

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -4333,7 +4333,7 @@ listFoldAnyDirectionChecks foldOperationName checkInfo =
                             if initialNumber == Just (Basics.toFloat operation.identity) then
                                 fixWith
                                     [ Fix.replaceRangeBy
-                                        { start = checkInfo.parentRange.start
+                                        { start = checkInfo.fnRange.start
                                         , end = (Node.range initialArgument).end
                                         }
                                         ("List." ++ operation.list)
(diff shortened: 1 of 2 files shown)


jfmengels (approved):
Thank you for fixing this @lue-bird :pray:

PR #88, opened 2023-03-04
Cache ImportLookup

lue-bird:
Following up #84, fixing small things and adding a context field for a a potentially already cashed `ImportLookup`


jfmengels:
All good changes :+1:

PR #89, opened 2023-03-04
Compute import lookup once after imports have been visited

jfmengels:
I think we can get rid of the "maybe cache".
It is possible that to support the `let` variables, we may need to touch that import lookup when switching scopes in the project and need to recompute the import lookup somewhat, but we should avoid to recompute the things I moved to `afterImportsVisitor` anyway.


lue-bird (approved):
:+1:

>touch that import lookup when switching scopes in the project

In my current design, the binding info isn't part of the `ImportLookup` so this should be fine
```elm
type alias QualifyResources a =
    { a
        | importLookup : ImportLookup
        , moduleBindings : Set String
        , localBindings : RangeDict (Set String)
    }
```

PR #90, opened 2023-03-05
Track bindings in scope

lue-bird:
- [x] value, function declaration bindings
- [x] declaration argument list bindings
- [x] variant bindings
- [x] let declaration, destructuring bindings
- [x] let declaration argument list bindings
- [x] lambda argument list bindings
- [x] case argument bindings
- [x] reset local bindings on declaration visit
- [x] remove local ranges after patterns on expression exit visit
- [x] tests for all mentioned binding kinds

Nice to have the import shadowing edge cases covered

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -847,6 +847,7 @@ type alias ModuleContext =
     , exposedVariantTypes : Exposed
     , moduleBindings : Set String
     , localBindings : RangeDict (Set String)
+    , branchLocalBindings : RangeDict (Set String)
     , rangesToIgnore : RangeDict ()
     , rightSidesOfPlusPlus : RangeDict ()
     , customTypesToReportInCases : Set ( ModuleName, ConstructorName )
@@ -954,6 +955,7 @@ fromProjectToModule =
                     }
             , moduleBindings = Set.empty
             , localBindings = RangeDict.empty
+            , branchLocalBindings = RangeDict.empty
             , rangesToIgnore = RangeDict.empty
             , rightSidesOfPlusPlus = RangeDict.empty
             , localIgnoredCustomTypes = []
@@ -1117,7 +1119,7 @@ dependenciesVisitor typeNamesAsStrings dict context =
       else
         [ errorForUnknownIgnoredConstructor unknownTypesToIgnore ]
     , { customTypesToReportInCases = customTypesToReportInCases
-      , exposedVariants = Dict.union  dependencyExposedVariants context.exposedVariants
+      , exposedVariants = Dict.union dependencyExposedVariants context.exposedVariants
       }
     )
 
@@ -1170,7 +1172,7 @@ declarationVisitor declarationNode context =
                 in
                 { context
                     | exposedVariants =
-                        Set.union  variantNames context.exposedVariants
+                        Set.union variantNames context.exposedVariants
                 }
 
             else
@@ -1198,9 +1200,13 @@ declarationVisitor declarationNode context =
 expressionVisitor : Node Expression -> ModuleContext -> ( List (Error {}), ModuleContext )
 expressionVisitor node context =
     let
+        expressionRange : Range
+        expressionRange =
+            Node.range node
+
         contextWithInferredConstants : ModuleContext
         contextWithInferredConstants =
-            case RangeDict.get (Node.range node) context.inferredConstantsDict of
+            case RangeDict.get expressionRange context.inferredConstantsDict of
                 Nothing ->
                     context
 
@@ -1213,17 +1219,37 @@ expressionVisitor node context =
                         | inferredConstants = ( inferredConstants, previous :: previousStack )
                     }
     in
-    if RangeDict.member (Node.range node) context.rangesToIgnore then
+    if RangeDict.member expressionRange context.rangesToIgnore then
         ( [], contextWithInferredConstants )
 
     else
         let
+            withExpressionSurfaceBindings : RangeDict (Set String)
+            withExpressionSurfaceBindings =
+                RangeDict.insert expressionRange (expressionSurfaceBindings (Node.value node)) context.localBindings
+
+            withNewBranchLocalBindings : RangeDict (Set String)
+            withNewBranchLocalBindings =
+                RangeDict.union (expressionBranchLocalBindings (Node.value node))
+                    context.branchLocalBindings
+
             contextWithInferredConstantsAndLocalBindings : ModuleContext
             contextWithInferredConstantsAndLocalBindings =
-                { contextWithInferredConstants
-                    | localBindings =
-                        RangeDict.union (expressionSurfaceBindings node) context.localBindings
-                }
+                case RangeDict.get expressionRange context.branchLocalBindings of
+                    Nothing ->
+                        { contextWithInferredConstants
+                            | localBindings = withExpressionSurfaceBindings
+                            , branchLocalBindings =
+                                withNewBranchLocalBindings
+                        }
+
+                    Just currentBranchLocalBindings ->
+                        { contextWithInferredConstants
+                            | localBindings =
+                                RangeDict.insert expressionRange currentBranchLocalBindings withExpressionSurfaceBindings
+                            , branchLocalBindings =
+                                RangeDict.remove expressionRange withNewBranchLocalBindings
+                        }
 
             { errors, rangesToIgnore, rightSidesOfPlusPlus, inferredConstants } =
                 expressionVisitorHelp node contextWithInferredConstantsAndLocalBindings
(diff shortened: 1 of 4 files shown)


src/Simplify/RangeDict.elm
```diff
@@ -1,10 +1,11 @@
-module Simplify.RangeDict exposing (RangeDict, empty, get, insert, member)
+module Simplify.RangeDict exposing (RangeDict, any, diff, empty, fromList, get, insert, map, mapFromList, member, singleton, union)
 
 import Dict exposing (Dict)
 import Elm.Syntax.Range exposing (Range)
 
 
 type alias RangeDict v =
+    -- TODO: make opaque to disallow type confusion with other String dicts
     Dict String v
```
lue-bird, line 9: What's your opinion? The idea of `RangeDict` in general seems to match with [`miniBill/elm-generic-dict`](https://dark.elm.dmy.fr/packages/miniBill/elm-generic-dict/latest/)

src/Simplify/RangeDict.elm
```diff
@@ -13,19 +14,81 @@ empty =
...
+            )
+            associations
+        )
+
+
 insert : Range -> v -> RangeDict v -> RangeDict v
-insert range =
-    Dict.insert (rangeAsString range)
+insert range rangeDict =
+    Dict.insert (rangeAsString range) rangeDict
```
lue-bird, line 55: A certain someone curried their functions as well :smile:

src/Simplify.elm
```diff
@@ -846,26 +845,16 @@ type alias ModuleContext =
...
-            ModuleName
-            { alias : Maybe ModuleName
-            , exposed : Exposed
-            }
-    , rangesToIgnore : List Range
-    , rightSidesOfPlusPlus : List Range
+    , moduleBindings : Set String
+    , localBindings : RangeDict (Set String)
+    , rangesToIgnore : RangeDict ()
+    , rightSidesOfPlusPlus : RangeDict ()
```
lue-bird, line 851: Going from `List Range` to `RangeDict ()` is totally unrelated to tracking local bindings. It shouldn't hurt anyone either but if you want, I can revert it and maybe open in a separate PR

jfmengels: I think the idea when I made them into lists is that they're always going to be quite small, so a list would be fine in practice. I think it might be slightly slower with a RangeDict, but I'm fine with using it, since it's not going to impact the performance much if they're tiny collections anyway.

tests/SimplifyTest.elm
```diff
@@ -309,6 +309,297 @@ a = List.foldl f x << UniqueList.toList
...
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+import Set exposing (foldl)
+a =
+    let
+        doIt ( _, Node _ ({ foldl } :: _) ) =
+            ()
+        
+        doItBetter x =
+            Set.foldl f x
```
jfmengels, line 516: Should this not be `foldl f x`? `foldl` from `doIt` is not accessible here.

src/Simplify.elm
```diff
@@ -1154,100 +1198,158 @@ declarationVisitor declarationNode context =
...
+        (expressionSurfaceBindingIntroductions expression)
+
+
+{-| Whenever you add ranges on expression enter, the same ranges should be removed on expression exit.
+Having one function finding unique ranges and a function for extracting bindings there ensures said consistency.
+
+An alternative approach would be to use some kind of tree structure
+with parent and sub ranges and bindings as leaves (maybe a "trie", tho I've not seen one as an elm package).
+
+Removing all bindings for an expression's range on leave would then be trivial
```
jfmengels, line 1259: What I usually do is to have a stack of scopes, and a separate RangeDict to know when to pop the stack in the exit visitor.

Another possibility would be to have to store the elements collected by `expressionSurfaceBindings` and store them in a RangeDict. Whenever the expression corresponding to the larger scope is encountered, we can remove all what was contained.

jfmengels (commented):
Great work! I think there are a few issues in the results (as pointed in my comments).

Also sorry for the delay in reviewing this, I needed to recharge a little bit.
Also, I tend to review commit by commit and commit 1ee06c432a22c21826a88bba3e6db3b7c08ee9b8 was really hard to read, mostly because two things were done in the same commit (using `checkInfo.<field>`, and adding resources). For the next time, I would recommend you split these up into separate commits. I imagine you wanted to add the resources then noticed that it would be easier to do if everything was qualified with `checkInfo.`.

What I usually do when I notice something like this: I'm working on adding the resources, then I notice that things would be easier with the qualification. I'll then stash my changes, qualify everything, commit, and then pop the stash to resume the previous work which is now easier to do. I do this every time I notice something that would make my current work easier. It also makes the work a lot easier to review, because if I see a large commit where it says "Avoid destructuring checkInfo in all functions" and I see similar work, I can trust you did only that and skim through the changes. But if the two changes are done in a single commit, I have to carefully look at what changed and figure out what is changing the behavior and what is only refactoring/cosmetic.

Anyway, just a tip for next time. Not sure it will work for you, but it's worked well for me :slightly_smiling_face:

src/Simplify.elm
```diff
@@ -1207,24 +1230,118 @@ expressionVisitor node context =
         , { newContext
             | rangesToIgnore = rangesToIgnore ++ newContext.rangesToIgnore
             , rightSidesOfPlusPlus = rightSidesOfPlusPlus ++ newContext.rightSidesOfPlusPlus
-            , inferredConstantsDict = List.foldl (\( range, constants ) acc -> RangeDict.insert range constants acc) newContext.inferredConstantsDict inferredConstants
+            , inferredConstantsDict = RangeDict.union (RangeDict.fromList inferredConstants) newContext.inferredConstantsDict
```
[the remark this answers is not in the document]
lue-bird: I thought it solved the same problem in a simpler way using a helper that wasn't available at the time. Your explanation makes sense.

src/Simplify.elm
```diff
@@ -932,22 +929,40 @@ fromProjectToModule =
...
+                    List.foldl
+                        (\import_ importLookup ->
+                            let
+                                importInfo : { moduleName : ModuleName, exposed : Exposed, alias : Maybe ModuleName }
+                                importInfo =
+                                    importContext import_
+                            in
+                            insertImport importInfo.moduleName { alias = importInfo.alias, exposed = importInfo.exposed } importLookup
+                        )
+                        Dict.empty
```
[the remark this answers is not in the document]
lue-bird: >if someone has an import import Platform.Cmd exposing (none), then the Cmd alias will currently be forgotten about as well as the Cmd type.

Oh, what? That's so weird because for example
> import Basics

you can still have unqualified `identity` etc. Maybe a special case with `exposing (..)` or `Basics`?

tests/SimplifyTest.elm
```diff
@@ -309,6 +309,297 @@ a = List.foldl f x << UniqueList.toList
...
+                            , under = "List.foldl"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+import Set exposing (foldl)
+a =
+    let
+        doIt ( _, Node _ ({ foldl } :: _) ) =
+            ()
+    in
+    Set.foldl f x
```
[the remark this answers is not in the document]
lue-bird: no? `foldl` is an argument to a let function, not destructured

src/Simplify.elm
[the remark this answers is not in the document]
jfmengels: Sorry, I meant to say that with the current implementation (but I might be wrong, would probably warrant a test), if we see a `import Platform.Cmd exposing (none)`, then `Cmd.none` is not going to be known, because the prelude import got overriden because of the `Dict.union`. The `insertImport` approach handles that correctly.

(when I say "current", I meant the one from when I did the review, because I see you applied my suggestion now).

tests/SimplifyTest.elm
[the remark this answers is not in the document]
lue-bird: right

src/Simplify.elm
```diff
@@ -1198,42 +1198,42 @@ declarationVisitor declarationNode context =
...
         ( errors
-        , { newContext
-            | rangesToIgnore = rangesToIgnore ++ newContext.rangesToIgnore
-            , rightSidesOfPlusPlus = rightSidesOfPlusPlus ++ newContext.rightSidesOfPlusPlus
-            , inferredConstantsDict = RangeDict.union (RangeDict.fromList inferredConstants) newContext.inferredConstantsDict
+        , { contextWithInferredConstantsAndLocalBindings
+            | rangesToIgnore = rangesToIgnore ++ context.rangesToIgnore
+            , rightSidesOfPlusPlus = rightSidesOfPlusPlus ++ context.rightSidesOfPlusPlus
+            , inferredConstantsDict =
+                RangeDict.union context.inferredConstantsDict (RangeDict.fromList inferredConstants)
```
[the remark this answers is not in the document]
lue-bird: Found a couple more places like that. Now I know the right order at least

tests/SimplifyTest.elm
lue-bird: right

lue-bird:
Now we should be all set.
I'll try to keep commits more on the "do one obvious thing" side now.
In my mind, a code style change didn't seem to require a separate commit. However, keeping in mind that others review commit by commit, separating makes more sense yeah.

jfmengels (approved):
LGTM! I'll merge this now. This now enables #82, correct?

PR #92, opened 2023-03-11
Extract a variable for tests that use the default settings

jfmengels:
Makes the tests slightly simpler, but maybe this can speed them up too a little bit (not that it's strictly necessary)


lue-bird (approved):
nice

PR #93, opened 2023-03-11
Various performance speedups

jfmengels:

I can remove a few of these if you think that it makes it really hard to read, let me know what you think!

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1107,11 +1107,11 @@ dependenciesVisitor typeNamesAsStrings dict context =
 
         dependencyExposedVariants : Dict ModuleName (Set String)
         dependencyExposedVariants =
-            modules
-                |> List.map
-                    (\moduleDoc ->
-                        ( AstHelpers.moduleNameFromString moduleDoc.name
-                        , moduleDoc.unions
+            List.foldl
+                (\moduleDoc acc ->
+                    Dict.insert
+                        (AstHelpers.moduleNameFromString moduleDoc.name)
+                        (moduleDoc.unions
                             |> List.concatMap
                                 (\union ->
                                     union.tags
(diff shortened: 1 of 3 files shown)


src/Simplify.elm
```diff
@@ -1310,19 +1310,18 @@ expressionBranchLocalBindings expression =
...
-                                    )
-                                    soFar
+                            RangeDict.insert
+                                (Node.range (Node.value letFunctionOrValueDeclaration.declaration).expression)
+                                (AstHelpers.patternListBindings
+                                    (Node.value letFunctionOrValueDeclaration.declaration).arguments
+                                )
+                                acc
 
                         _ ->
```
lue-bird, line 1327: Just FYI: identity is curried, yes. but it's the base function so it doesn't come with performance costs

lue-bird (approved):
I bet there is a middle ground somewhere that doesn't downgrade readability this much, like having `setMapFromList`, `dictMapFromList` etc. but since you removed `RangeDict.mapFromList` I take it that you dislike this?

src/Simplify.elm
```diff
@@ -1310,19 +1310,18 @@ expressionBranchLocalBindings expression =
...
-                            identity
+                            acc
```
lue-bird, line 1328: Just FYI: identity is curried, yes. but it's the base function so it doesn't come with performance costs

jfmengels: The problem here is not `identity`, it's that the fold function is defined as taking a single argument, when it will be called with 2 in practice. It's therefore faster if we define it more explicitly with 2 arguments.

jfmengels:
> I bet there is a middle ground somewhere that doesn't downgrade readability this much, like having setMapFromList, dictMapFromList etc. but since you removed RangeDict.mapFromList I take it that you dislike this?

I think so as well, I'm thinking about adding functions like these and maybe making that into a package, who knows. I'm not against this, but often the function will also need to take a initial value. My gut feeling is that these might end up complex enough that an inline fold might be easier, but it's worth looking into.

Maybe I should re-add `RangeDict.mapFromList`, what do you think? I was initially hoping for more changes, that would have required this inlining, but decided to hold back on those, at least for now, so maybe I should?

jfmengels:
I've re-added `RangeDict.mapFromList`

PR #96, opened 2023-03-12
Add Dict.partition simplifications

lue-bird:
Simply uses `collectionPartitionChecks dictCollection`
```elm
Dict.partition f Dict.empty
--> ( Dict.empty, Dict.empty )

Dict.partition (always True) dict
--> ( dict, Dict.empty )

Dict.partition (always False) dict
--> ( Dict.empty, dict )

Dict.partition (always False)
--> (Tuple.pair Dict.empty)
```
only the first three are shown in the summary

### DIFF
--- a/tests/SimplifyTest.elm
+++ b/tests/SimplifyTest.elm
@@ -14701,6 +14701,7 @@ dictSimplificationTests =
         , dictToListTests
         , dictSizeTests
         , dictMemberTests
+        , dictPartitionTests
         ]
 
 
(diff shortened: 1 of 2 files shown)


tests/SimplifyTest.elm
```diff
@@ -14976,6 +14977,174 @@ a = False
...
+                            , under = "Dict.partition"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = ( x, Dict.empty )
+"""
+                        ]
+        , test "should not replace Dict.partition (always True)" <|
+            -- We'd likely need an anonymous function which could introduce naming conflicts
+            -- Could be improved if we knew what names are available at this point in scope (or are used anywhere)
+            -- so that we can generate a unique variable.
```
lue-bird, line 15074: We now have the means to create argument names that aren't locally shadowed. I don't really like the lambda fix, tho

jfmengels: Oh yeah, we could do that now :eyes: 
The naming convention we choose could displease people, but as a suggestion, it could definitely be a good start (and that's partially why there are prompts for fixes).

jfmengels (approved):
Nice! I was looking at the tests first, and then wanted to look for the implementation, and there was barely any code and was confused for a bit :sweat_smile:

PR #97, opened 2023-03-12
Add Result.toMaybe simplifications

lue-bird:
```elm
Result.toMaybe (Ok x)
--> Just x

Result.toMaybe (Err e)
--> Nothing

Result.toMaybe << Err
--> always Nothing

Result.toMaybe << Ok
--> Just
```
Only the first two are shown in the summary


jfmengels (approved):
Nice 😊

PR #100, opened 2023-03-13
Add ability to reduce lambdas

lue-bird:
Adds a simple algorithm that tries to reduce
```elm
(\f -> List.map f) --> List.map
(\f -> \xs -> List.map f xs) --> List.map
(\f xs -> (f |> List.map) <| xs) --> List.map
```
This PR currently only uses these reducing function parsers for `Result.mapError`, but this should be useful in a lot of places.

Reducing only looks at lambdas, so it isn't "fully complete" since it doesn't reduce stuff like case-ofs or let blocks, e.g.
```elm
case unrelated of
    A -> \a -> f a
    B -> \b -> f b
```
or
```elm
case unrelated of
    A -> \a b -> function b
    B -> \a -> function
```
(_maybe_ that could be a separate check. Moving common arguments to an upper level might be too opinionated. A better place might be [`elm-review-reducible-lambdas`](https://package.elm-lang.org/packages/jsuder-xx/elm-review-reducible-lambdas/latest/NoEtaReducibleLambdas))

### DIFF
--- a/src/Simplify/AstHelpers.elm
+++ b/src/Simplify/AstHelpers.elm
@@ -365,6 +365,89 @@ isIdentity lookupTable baseNode =
             False
 
 
+getReducedLambdaToCall :
+    Node Expression
+    ->
+        Maybe
+            { nodeRange : Range
+            , fnName : String
+            , fnRange : Range
+            , callArguments : List (Node Expression)
+            , lambdaPatterns : List (Node Pattern)
+            }
+getReducedLambdaToCall expressionNode =
+    -- maybe a version of this is better located in Normalize?
+    case getCollapsedLambda expressionNode of
+        Just lambda ->
+            case getCollapsedValueOrFunction lambda.expression of
+                Just call ->
+                    let
+                        ( reducedCallArguments, reducedLambdaPatterns ) =
+                            drop2EndingsWhile
+                                (\( argument, pattern ) ->
+                                    case Node.value (removeParens argument) of
+                                        Expression.FunctionOrValue [] argument0Name ->
+                                            case getVarPattern pattern of
+                                                Just pattern0Name ->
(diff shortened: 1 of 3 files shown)


src/Simplify/AstHelpers.elm
```diff
@@ -306,6 +367,228 @@ isIdentity lookupTable baseNode =
...
+    ->
+        Maybe
+            { nodeRange : Range
+            , fnName : String
+            , fnRange : Range
+            , callArguments : List (Node Expression)
+            , lambdaPatterns : List (Node Pattern)
+            }
+getReducedLambdaToCall expressionNode =
+    -- maybe a version of this is better located in Normalize?
```
jfmengels, line 499: I'd say so. also, we might want to have this logic as well in the Normalize logic, so that we notice that `f Err == f (\a -> Err a)`. Let's do this in a separate PR

tests/SimplifyTest.elm
```diff
@@ -673,6 +674,132 @@ a =
...
+"""
+                    |> Review.Test.run ruleWithDefaults
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Using Result.mapError on Err will result in Err with the function applied to the error"
+                            , details = [ "You can replace this call by Err with the function directly applied to the error itself." ]
+                            , under = "Result.mapError"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = Err << f
```
jfmengels, line 696: The only part I'm unsure about is as to whether we should keep the user's original code for the error function. But... I think this is going to be okay in practice?

jfmengels (approved):
Really nice! I imagine we have plenty of places where we can start applying this!

PR #102, opened 2023-03-13
Add simplifications for Dict.toList, then List.map tuple part access

lue-bird:
As shown in the Dict section of https://github.com/jfmengels/elm-review-simplify/issues/2

```elm
List.map Tuple.first (Dict.toList dict)
--> Dict.keys dict
List.map Tuple.second (Dict.toList dict)
--> Dict.values dict

List.map (\( part0, _ ) -> part0) (Dict.toList dict)
--> Dict.keys dict
List.map (\( part1, _ ) -> part1) (Dict.toList dict)
--> Dict.values dict

List.map Tuple.first << Dict.toList
--> Dict.keys
List.map Tuple.second << Dict.toList
--> Dict.values
```
Only the first 2 are shown in the summary

What might be interesting to add is
```elm
List.map (Tuple.first >> f) (Dict.fromList dict)
--> List.map f (Dict.keys dict)
List.map (\( part0, _ ) -> f part0) (Dict.fromList dict)
--> List.map f (Dict.keys dict)
```
etc

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -2118,7 +2118,7 @@ functionCallChecks =
         , ( ( [ "List" ], "head" ), listHeadChecks )
         , ( ( [ "List" ], "tail" ), listTailChecks )
         , ( ( [ "List" ], "member" ), listMemberChecks )
-        , ( ( [ "List" ], "map" ), collectionMapChecks listCollection )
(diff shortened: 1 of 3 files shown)


jfmengels (approved):
Nice! This caught a few places on my work project, including one instance using composition 💪

PR #105, opened 2023-04-17
Fix #104

### LINKED ISSUE #104: `List.append` bug – keeps adding more and more lines
Culprit code:

```elm
list =
    [ 1 ]
        |> List.append [ 2 ]
```

image

image

SSCCE: https://github.com/lydell/elm-review-simplify-issue

lue-bird:
Corrects #104 

The error was that I blindly copied the range replacing from `(++)` which was wrong because the order: "first, then second argument" cannot be guaranteed

There is no way however to fix this without being able to extract source code of the piped list, therefore no fixes will be provided.

### DIFF
--- a/tests/SimplifyTest.elm
+++ b/tests/SimplifyTest.elm
@@ -6333,7 +6333,7 @@ d = List.append xs [ 1 ]
 """
                     |> Review.Test.run ruleWithDefaults
                     |> Review.Test.expectNoErrors
-        , test "should report List.append on two list literals" <|
+        , test "should report List.append applied on two list literals" <|
             \() ->
                 """module A exposing (..)
 a = List.append [b] [c,d,0]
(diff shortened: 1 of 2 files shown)


tests/SimplifyTest.elm
```diff
@@ -6349,6 +6349,35 @@ a = List.append [b] [c,d,0]
...
+a = [c,d,0] |> List.append [b]
+"""
+                    |> Review.Test.run ruleWithDefaults
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Appending literal lists could be simplified to be a single List"
+                            , details = [ "Try moving all the elements into a single list." ]
+                            , under = "List.append"
+                            }
+                        ]
```
jfmengels, line 6380: Why do we not have a fix for this one? :thinking:

lue-bird: I describe this in the PR body:
The elements that should be later in the combined list are located before when using `|>`, so
> There is no way to fix this without being able to extract source code of the piped list, therefore no fixes will be provided.

jfmengels: Ah sorry, my bad, I didn't read it, went straight for the commits. But we do have a mechanism for that :)
https://package.elm-lang.org/packages/jfmengels/elm-review/latest/Review-Rule#withSourceCodeExtractor

We already use that, so that is the `extractSourceCode` field in the `Context`., You can grab the code in between the brackets and put them at the end.

jfmengels: Yes, let's add the `sourceCodeExtractor` to the CheckInfo, that sounds good :+1:

jfmengels:
Thank you for jumping on the bug @lue-bird :pray:  :balloon:

PR #106, opened 2023-04-17
Add source code extractor to checkInfo

lue-bird:
Also adds a fix and test for `some |> List.append some`


jfmengels:
Nice :grin:

PR #117, opened 2023-08-13
Simplify List.any ((==) x) to List.member x

### LINKED ISSUE #110: Simplify `List.any ((==) x)` into `List.member x`
I found a bunch of  `List.any ((==) x)` which I could replace with `List.member x`. I think this would be nice to always simplify!

jfmengels:
Fixes #110

(cc @lydell)


lydell:
Awesome, thank you!

src/Simplify.elm
```diff
@@ -5307,7 +5310,20 @@ listAnyChecks checkInfo =
...
                 _ ->
-                    []
+                    case Evaluate.isEqualToSomethingFunction checkInfo.firstArg of
+                        Nothing ->
+                            []
+
+                        Just rangesToRemove ->
+                            [ Rule.errorWithFix
+                                { message = "Use List.member instead"
+                                , details = [ "This call to List.any checks for the presence of a value, which what List.member is for." ]
```
lue-bird, line 5320: some words are missing in "[...], which what List.member is for."
If I had a guess, you probably wanted "which is what List.member is for" or "This call [...] checks for [...]. List.member is meant for this exact purpose."

PR #118, opened 2023-08-14
correctly fix List.concat of only list literals in pipeline

lue-bird:
Whenever we provided a fix for `List.concat [ [ ...], ..., [ ... ] ]`, we merged the elements and removed the `List.concat` function reference.

So whenever there was a pipeline like `List.concat <| [[1,2,3],[4,5,6]]`, the fix source code would not compile:

    I was unable to parse the source code after applying the fixes. Here is
    the result of the automatic fixing:
    
      ```
        module A exposing (..)
        a =  <| [1,2,3,4,5,6]
    
      ```
    
    This is problematic because fixes are meant to help the user, and applying
    this fix will give them more work to do. After the fix has been applied,
    the problem should be solved and the user should not have to think about it
    anymore. If a fix can not be applied fully, it should not be applied at
    all.

The fix in this PR uses `keepOnlyFix (range of first arg)` to solve this and adds tests to verify it now works.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -4171,8 +4171,8 @@ listConcatChecks checkInfo =
                             , details = [ "Try moving all the elements into a single list." ]
                             }
                             checkInfo.parentRange
-                            (Fix.removeRange checkInfo.fnRange
-                                :: List.concatMap removeBoundariesFix args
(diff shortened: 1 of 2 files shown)


lue-bird:
reopening from a different branch because something went wrong here somehow

jfmengels:
Feel free to push `--force` a branch if something is broken (and if you're uncomfortable with this or another aspect of Git feel free to ask me for advice :+1: )

PR #119, opened 2023-08-14
correctly fix List.concat of only list literals in pipeline

lue-bird:
Whenever we currently provide a fix for `List.concat [ [ ...], ..., [ ... ] ]`, we merge the elements and remove the `List.concat` function reference.

So whenever there was a pipeline like `List.concat <| [[1,2,3],[4,5,6]]`, the fix source code would not compile:

    I was unable to parse the source code after applying the fixes. Here is
    the result of the automatic fixing:
    
      ```
        module A exposing (..)
        a =  <| [1,2,3,4,5,6]
    
      ```
    
    This is problematic because fixes are meant to help the user, and applying
    this fix will give them more work to do. After the fix has been applied,
    the problem should be solved and the user should not have to think about it
    anymore. If a fix can not be applied fully, it should not be applied at
    all.

The fix in this PR uses `keepOnlyFix` with the range of list arg to solve this and adds tests to verify pipelines work correctly now.

One extra thing this PR contains which is only slighly related: For `concatMap f [ one ]` the tests had wrong names and were missing some cases. I couldn't let them sit there :)

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -4208,8 +4208,8 @@ listConcatChecks checkInfo =
                                     , details = [ "Try moving all the elements into a single list." ]
                                     }
                                     checkInfo.parentRange
-                                    (Fix.removeRange checkInfo.fnRange
-                                        :: List.concatMap removeBoundariesFix args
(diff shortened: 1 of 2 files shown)


jfmengels (approved):
Very nice. I like how this improves the fix for the `List.concat [ ... ]` case too :)

Issue #41, opened 2022-08-20
Simplify `a - a` to `0`


miniBill:
**What the rule should do:**
Simplify `a-a` to `0`

**Example of things the rule would report:**

```elm
item.quality - item.quality
--> 0
```

**Example of things the rule would not report:**

```elm
item.quality - (.quality item)
```

jfmengels:
> item.quality - (.quality item)

Why would you not want this be reported? Except for the fact that it can be hard to detect (which may actually not be all that complicated).

I also think that we could report the following
```elm
item.quality + (-item.quality)
--> 0
```


And maybe we could also simplify things like (more work though)

```elm
item.quality + 3 - item.quality + n
--> 3 + n
```

though I don't know if this would create for annoyances in how you want to write your code.

miniBill:
It would be reported after the `.quality item -> item.quality` transformation I guess?

`a + (-a)` should be fine

`a + 3 - a + n` would be more work

While we're at it, does it make sense to simplify

a) `(a + 1) + 1`
b) `a + 1 + 1
?

jfmengels:
I don't think we have the `.quality item -> item.quality` transformation :thinking: 

For a) and b), do you mean to transform them to `a + 2`? I would say no, because that can be related to how you want to write code.

The example I usually have in mind is something like:

```elm
Process.sleep (3 * 60 * 1000)
```
Abole it becomes easy to see how long to wait: 1000 milliseconds is one second, times 60 is one minute, times 3 is 3 minutes.
We could simplify it to
```elm
Process.sleep 180000
```
but that would be a lot harder to read and feel like a much more arbitrary/magic value compared to the broken-down version we had before.

miniBill:
Agreed!

lue-bird:
@ future implementor: As with #58, these simplifications aren't always correct with `Float`
```elm
> (0 / 0) - (0 / 0)
NaN : Float
> (1 / 0) - (1 / 0)
NaN : Float
> -(1 / 0) - -(1 / 0)
NaN : Float
```

miniBill:
I'm closing this in favor of creating https://github.com/miniBill/elm-review-optimize

jfmengels:
I'll re-open this one, because I can still see this one being valuable in Simplify :)

miniBill:
Considering what Lue said, I'm guessing that Simplify would have the rule but not the simplification, whereas Optimize could have both?

jfmengels:
[...]

PR #120, opened 2023-08-14
Add expectNaN configuration option

jfmengels:
New opt-in configuration option `expectNaN` which will disable some simplifications when the user indicates their
project is likely to use `NaN` values. This disables the following simplifications:
- `x == x` to `True`
- `List.member x [ x ]` to `True`
- `n * 0` to `0`

When `expectNaN` is not used, `n * 0` to `0` will be reported (as before) but will now have an autofix.

I may have missed other places where `NaN` could be a problem. This will enable working on https://github.com/jfmengels/elm-review-simplify/issues/41 safely.

PR can be reviewed commit by commit.


src/Simplify.elm
```diff
@@ -973,8 +1021,8 @@ fromModuleToProject =
         )
 
 
-fromProjectToModule : Rule.ContextCreator ProjectContext ModuleContext
-fromProjectToModule =
+fromProjectToModule : Bool -> Rule.ContextCreator ProjectContext ModuleContext
```
lue-bird, line 1024: I'd just have this argument be `{ expectNaN : Bool }` or even `{ config_ | expectNaN : Bool }`

src/Simplify.elm
```diff
@@ -846,7 +847,10 @@ type Configuration
 -}
 defaults : Configuration
 defaults =
-    Configuration { ignoreConstructors = [] }
+    Configuration
+        { ignoreConstructors = []
+        , expectNaN = False
```
lue-bird, line 852: This is I guess what you wanted feedback on?
It's a really hard decision because we want to balance trustable, bug-free refactorings with convenience.
I like that users can decide to tell simplify that they don't expect NaN and get the convenience. NaN is used almost never so the trade-off to require explicit confirmation that NaNs are used makes sense.

What we do need in my opinion is a very clear indication in the documentation that you should decide whether you want to set this option for your whole project.
(If I was writing the rule, I'd probably make the convenience opt-in, so that users always explicitly know of the trade-offs and consent to them but good documentation should do the job just as well.)

src/Simplify.elm
```diff
@@ -879,7 +883,50 @@ Please let me know by opening an issue if you do use this function, I am very cu
 -}
 ignoreCaseOfForTypes : List String -> Configuration -> Configuration
 ignoreCaseOfForTypes ignoreConstructors (Configuration config) =
-    Configuration { config | ignoreConstructors = ignoreConstructors ++ config.ignoreConstructors }
+    Configuration { ignoreConstructors = ignoreConstructors ++ config.ignoreConstructors, expectNaN = config.expectNaN }
+
+
+{-| This rule will apply simplifications that are always safe to apply without risk of behavior change.
+However, some rare changes can actually impact behavior when encountering
+[`NaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN) values.
```
lue-bird, line 891: To me this reads very odd. Maybe
```md
Usually, `elm-review-simplify` will only suggest simplifications that are safe to apply without risk of changing the original behavior.
However, when encountering
[`NaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN) values, some simplifications can actually impact behavior.
```

lue-bird (approved):

lue-bird:
For all the errors that can't be fixed because of the possibility of `NaN`, it might make sense to add a reminder that `|> Simplify.expectNaN` is configured and that a fix is only suggested when this setting isn't used.

jfmengels:
> For all the errors that don't provide fixes because of the possibility of NaN, it might make sense to add a reminder that the `|> Simplify.expectNaN` configuration is active and that a fix is only suggested when this setting isn't added.

If I'm not mistaken, for all the places that care about `NaN` the error is disabled, not just the fix. But I agree that we should probably mention the configuration if we only skip the fix.

Issue #46, opened 2022-08-27
The fix removed a useful comment


lydell:
The rule simplifed some list code:

```diff
                 (DesignSystem.fontColor Blue500
-                    :: Element.Font.underline
-                    -- This class is used in app.ts.
-                    :: Element.htmlAttribute (Html.Attributes.class "js-ReloadPage")
-                    :: [ DesignSystem.fontWeight DesignSystem.Bold ]
+                    :: [ Element.Font.underline, Element.htmlAttribute (Html.Attributes.class "js-ReloadPage"), DesignSystem.fontWeight DesignSystem.Bold ]
                 )
```

But I lost that `-- This class is used in app.ts.` comment.

Thoughts?

jfmengels:
Yeah, this is unfortunate, but it's really hard to handle comments at the moment, and I don't know of a design that would make comments work out better.

I see 3 ways of improving this:
1. Have `elm-syntax` better attach the comments to the AST. Unfortunately I have no clue how this could be achieved, or what "better attached" would mean in practice.
2. Have `elm-review` report a warning whenever it notices that comments have changed after a fix, and show that in the fix prompt.
3. In this specific instance, be a bit more conservative on the fix. I thin we're basically replacing the from `Blue500` to `Element.Font.underline` by `[`, and we could try to only replace `::`, which may help with this case. But `elm-syntax` doesn't currently provide the range information about the operator, hence why this removes more than necessary.

In the meantime, this is what the fix prompt is for, so that you can re-add the comment or apply the fix manually yourself. So I'm happy that you were able to figure this out (at some point in the prompt or in the `git diff`). I would love to solve it, but even the easiest solution would require a major version of `elm-syntax` (which we will focus on one day).

PR #121, opened 2023-08-14
Enable access to the operator Range

lue-bird:
`elm-syntax` only gives us the operator symbol String but having access to the operator Range as well is valuable:
- we can replace only the operator instead of replacing everything between the operands, preserving comments and whitespace. → fixes https://github.com/jfmengels/elm-review-simplify/issues/46
- we can have a better error Range for operation simplification. Error ranges can quickly go out of hand which is the reason all function simplifications put the error range on the `fnRange`, not one of the arguments. Now we can do the same with operations

It seems this will be part of `elm-syntax` v8 https://github.com/stil4m/elm-syntax/issues/74 but now we can patch things up in the meantime :white_flower:

If the PR's solution seems too unstable and you want to wait for v8, I'm happy to wait as well.

Now what's included in this PR?
  - finding the operator's range (skips searching in all comments) and saving it in the `OperatorCheckInfo`
  - a more conservative fix for `:: [ ... ]` as hinted at in https://github.com/jfmengels/elm-review-simplify/issues/46
  - tests

That means that if this PR would be merged, we still need to change the other error ranges to the operator ranges.

### DIFF
--- a/tests/SimplifyTest.elm
+++ b/tests/SimplifyTest.elm
@@ -17398,7 +17398,7 @@ a =
         |> h
 """
                         ]
-        , test "should replace >> when used directly in a |> pipeline (with parentheses)" <|
+        , test "should replace >> when used directly in a |> pipeline (right argument >> inside parentheses)" <|
             \() ->
                 """module A exposing (..)
 a =
@@ -17419,6 +17419,29 @@ a =
 a =
     b
         |> f |> g
+"""
+                        ]
+        , test "should replace >> when used directly in a |> pipeline (right argument |> >> inside parentheses)" <|
+            \() ->
+                """module A exposing (..)
+a =
+    b
+        |> (f |> g >> h)
+"""
+                    |> Review.Test.run ruleWithDefaults
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Use |> instead of >>"
+                            , details =
+                                [ "Because of the precedence of operators, using >> at this location is the same as using |>."
+                                , "Please use |> instead as that is more idiomatic in Elm and generally easier to read."
+                                ]
+                            , under = ">>"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a =
+    b
+        |> (f |> g |> h)
 """
                         ]
         , test "should replace << when used directly in a <| pipeline" <|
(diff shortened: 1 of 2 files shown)


tests/SimplifyTest.elm
```diff
@@ -12,6 +12,7 @@ all =
         [ configurationTests
         , qualifyTests
         , lambdaReduceTests
+        , operatorRangeTests
```
jfmengels, line 15: Maybe move this to the bottom of the list, since this is more of a test about the implementation than about a specific transformation.
I'd also be happy with moving the new tests inline next to the tests for `"Element added to the beginning of the list could be included in the list"`. That way when v8 comes around there won't be a good reason to have a dedicated describe.

jfmengels:
> I strongly wish this will be part of elm-syntax v8 but now we can patch things up in the meantime (should I add an issue in elm-syntax?)

Yes, same here. It will happen one day :wink: 

---

I'm a bit surprised about this PR, because I **just** merged the same thing in https://github.com/jfmengels/elm-review-simplify/pull/116/files#diff-6239bccafe2b75c7e560542732acf25d142d08608e1e0f0ffe9b53c09cfb1e0eR6540-R6545

I think we're doing the same thing, except that you're also handling comments, which is a good idea. Can you confirm that we're doing the same thing, or am I misunderstanding something?
If we are, then can you try replacing my code by your version? Let's go with your version since it's going to be more correct.

lue-bird:
Yeah, same thing, just more complicated. Btw `String.split "\n" --> String.lines` would be a nice one to add ;)

jfmengels (approved):
Alright, feel free to merge whenever you feel like it (just avoid a merge commit please :pray: ).

lue-bird:
>try replacing my code by your version

Ok, I think we're good to go

lue-bird:
> merge whenever you feel like it (just avoid a merge commit please :pray:).

before I mess something up, I assume you want "rebase and merge"?
image

jfmengels:
Yes (You do have some failing tests now though)

lue-bird:
>You do have some failing tests now though

Not sure I follow, the github actions says the tests pass

jfmengels:
Huh weird, it showed as failing to me yesterday. Alright go ahead then 😊

PR #122, opened 2023-08-14
Replace n - n by 0

jfmengels:
Fixes https://github.com/jfmengels/elm-review-simplify/issues/41

```
n - n
--> 0

-n + n
--> 0
```

I think we could do something nicer like

```elm
something + n + else - n -- notice the 2 n cancel each other out
--> something + else
```

but I'm leaving that to others/later.


src/Simplify.elm
```diff
@@ -2453,6 +2465,29 @@ plusChecks checkInfo =
...
+addingOppositesCheck checkInfo =
+    if checkInfo.expectNaN then
+        []
+
+    else
+        case Normalize.compare checkInfo checkInfo.left (Node Range.emptyRange (Expression.Negation checkInfo.right)) of
+            Normalize.ConfirmedEquality ->
+                [ Rule.errorWithFix
+                    { message = "Addition always results in 0"
+                    , details = [ "Opposite values are on both end of `-` which will always result in 0. You can replace the expression by 0." ]
```
lue-bird, line 2478: sounds weird.
Maybe
> These two values have an equal absolute value but an opposite sign. This means adding them they will cancel out to 0

PR #124, opened 2023-08-15
Report immediately invoked anonymous functions

### LINKED ISSUE #111: Simplify `x |> (\y -> f y)` into `f x`
I had some code like this:

```elm
example =
    [ Just label
    , info
    ]
        |> List.filterMap identity
        |> List.concatMap
            (\template ->
                template.children
            )

```

During a refactor, I removed `info` from the list, which was correct and compiles.

```diff
 example =
     [ Just label
-    , info
     ]
         |> List.filterMap identity
         |> List.concatMap
             (\template ->
                 template.children
             )
 
```

elm-review nicely notices that the `filterMap` and `concatMap` are unnecessary now:

```elm
example =
    label
        |> (\template ->
                template.children
           )
```

However, piping a variable to a lambda is unnecessary too. I could simplify it to:

```elm
example =
    label.children
```

After that, I happened to stumble on the same pattern in the wild:

- https://github.com/dillonkearns/elm-markdown/blob/7.0.1/src/Markdown/Renderer.elm#L370-L372
- https://github.com/dillonkearns/elm-markdown/blob/7.0.1/src/Markdown/Renderer.elm#L404-L405

So I think there’s potential for more simplifications here!

I think the pattern we are looking for is:

- A simple identifier (lets call it `x`)
- Piped to a lambda
- Whose first pattern is also a simple identifier (lets call it `y`)

Then we can get rid of the lambda and substitute `y` into `x`.

Examples:

- `x |> (\y -> f y)` -> `f x`
- `x |> (\y -> y.p)` -> `x.p`
- `x |> (\y z -> y + z)` -> `\z -> x + z`

What do you think?

jfmengels:
Fixes #111

This reports but does not autofix anonymous functions that get directly invoked.

I'd love feedback on this. The message and details are more important than for other errors because there is no autofix. Let me know if you have any feedback (on this aspect or another) cc @lydell. PR can be reviewed commit by commit.


src/Simplify.elm
```diff
@@ -1554,55 +1557,13 @@ expressionVisitorHelp node context =
         --  APPLIED LAMBDA FUNCTIONS --
         -------------------------------
         Expression.Application ((Node _ (Expression.ParenthesizedExpression (Node lambdaRange (Expression.LambdaExpression lambda)))) :: firstArgument :: _) ->
```
lue-bird, line 1559: The fact that this is only reported on direct apply, not `|>` or `<|` avoids many intentional situations like
```elm
pieceBodies
    |> List.map .data
    |> (\pieces -> newPiecesThisRound ++ pieces)
    |> takeFormations
    |> toBodies
```
so this rule gets a :+1: from me

src/Simplify.elm
```diff
@@ -7706,6 +7667,69 @@ introducesVariableOrUsesTypeConstructor context nodesToLookAt =
...
+                    secondPattern :: _ ->
+                        [ Fix.removeRange { start = allRange.start, end = (Node.range secondPattern).start }
+                        , Fix.removeRange (Node.range firstArgument)
+                        ]
+                )
+            ]
+
+        _ ->
+            [ Rule.error
+                { message = "Found immediately invoked anonymous function"
```
lue-bird, line 7719: You do this in a few error messages: "Found ...". why not just say what you found? Like
> Anonymous function is immediately invoked

lue-bird (approved):

src/Simplify.elm
jfmengels: Ah, I actually forgot about these. I guess I kind of want to support them to a certain extent (especially since this is not even resolving @lydell's original issue...) but I agree that we don't want to report an issue for the example you just showed.

Maybe the heuristic should be that if we pipe an expression to the anonymous function, then we only report it if the argument is not a pipeline itself.

```elm
-- Reported
pieceBodies
    |> (\pieces -> newPiecesThisRound ++ pieces)
    |> f

-- NOT reported
pieceBodies
    |> List.map .data
    |> (\pieces -> newPiecesThisRound ++ pieces)
```

And same thing for `<|`. What do you think?

src/Simplify.elm
jfmengels: Yeah, I keep getting confused about what I should say especially for `Simplify` (there are so many errors that there's not even a good template I feel like I could follow :sweat_smile: )

Your suggestion sounds really good :+1:

src/Simplify.elm
lydell: If this reports only `(\b -> b * 2) a` and not `a |> (\b -> b * 2)`, then it is a useful simplification, but it won’t solve the problem I reported. I don’t think I’ve ever encountered `(\b -> b * 2) a` in real code.

I do however have encountered `a |> (\b -> b * 2)`, in two types of situations:

1. Because the Simplify rule simplified a pipeline into that (like I mentioned in #111).
2. Because colleagues go into “pipeline mode” and forget that let-in exists. See below.

I personally _always_ dislike piping into a lambda, but I can agree on it in @lue-bird’s example with a longer pipeline. However, I don’t think it makes sense to _start_ a pipeline with a lambda.

```elm
bad1 =
    createFruitBasket seeds prices recommendations
        |> (\fruitBasket ->
                { fruits = fruitBasket.fruits
                , size = fruitBasket.sizes |> List.sum
                }
           )


good1 =
    let
        fruitBasket =
            createFruitBasket seeds prices recommendations
    in
    { fruits = fruitBasket.fruits
    , size = fruitBasket.sizes |> List.sum
    }


bad2 =
    createFruitBasket seeds prices recommendations
        |> (\fruitBasket -> fruitBasket.apples ++ fruitBasket.oranges)
        |> List.map viewFruit


good2 =
    let
        fruitBasket =
            createFruitBasket seeds prices recommendations
    in
    fruitBasket.apples
        ++ fruitBasket.oranges
        |> List.map viewFruit
```

jfmengels: I think it *can* make sense to wrap in a lambda is something a bit complex, like an `if` condition.

```elm
bad1 =
    createFruitBasket seeds prices recommendations
        |> (\fruitBasket ->
              if condition then
                Just { fruits = fruitBasket.fruits
                        , size = fruitBasket.sizes |> List.sum
                        }
             else
                Nothing
           )
```

though I would personally always recommend using a `let` variable instead.

But I feel like my proposal follows @lydell's examples, so let's go with that. I'll probably do this in a separate PR though.

lue-bird: You _might_ be able to make a case that more complex arguments are fine to be piped into a lambda
```elm
piecesExtractFromBodies
    { physicsBodies = model.physicsBodies
    , activeCombineStrategy = model.combineStrategy
    , countCombinedAsSeparate = True
    }
    |> (\pieces -> newPiecesThisRound ++ pieces)
    |> ...continues
```
My initial reaction was for this rule to only report with "simple" arguments (literals and function/value references) for pipes to be reported.
But now I think your suggestion will probably be a good rule and above style isn't all that great :+1:

jfmengels:
I added pipeline support to this branch in the way that was mentioned above.

You can try out the change using:
```
elm-review --template jfmengels/elm-review-simplify/preview#direct-lambda-calls
```

Let me know what you think!

lydell:
Awesome!

It found 8 errors at work, and all of them became better when rewritten, either by switching to let-in or by extracting a function.

jfmengels:
Awesome, then it's time to merge :blush:

PR #130, opened 2023-08-15
WIP: Cleanup

lue-bird:
I want to `elm/random` ones next, but first a bit of cleanup. If you don't like certain changes, let's discuss and use one style consistently.
#### done
- [x] de-duplicate `Node` access calls and prefer desctructuring
- [x] use `keepOnlyFix` instead where possible

### concrete TODOs
- [ ] stringReplaceChecks secondArg_, thirdArg_ → replacementStringArg, baseStringArg
- [ ] listTailChecks, listHeadChecks, stringJoinChecks, listAllChecks, listAnyChecks, stringReplaceChecks, listFoldAnyDirectionChecks → first that reports err
- [ ] listAllChecks: needs maybeListArg
- [ ] listFoldAnyDirectionChecks: initialArgument → initialArg
- [ ] subAndCmdBatchChecks: use neighborMap
- [ ] removeRecordFields: use consecutiveMap

I remember that you disliked when PRs contain merges, so I could use some help of what the alternatives are


src/Simplify.elm
```diff
@@ -1545,83 +1545,52 @@ expressionVisitorHelp (Node expressionRange expression) context =
...
-                    { lambdaRange = lambdaRange
-                    , lambda = lambda
-                    , firstArgument = firstArgument
-                    }
-                )
+        -----------------
+        -- APPLICATION --
+        -----------------
+        Expression.Application (applied :: applicationFirstArg :: applicationArgsAfterFirst) ->
+            case ( applied, ( applicationFirstArg, applicationArgsAfterFirst ) ) of
```
jfmengels, line 1552: We could pattern match on a triplet here instead of nested tuples. That said, I don't think you need to pattern match on `applicationFirstArg`, in this case it's only used to give it a different name.

And I'm even unsure as to whether we need to pattern match on `applicationArgsAfterFirst`, I'd probably do a new case expression in the `PrefixOperator` case, that way this case expression is only on `applied` and is a bit simpler. But feel free to do what you prefer for this one :+1:

jfmengels (approved):
All great changes, and thanks for splitting this up in plenty of tiny and focused commits :+1: 

In my opinion you can merge this all already and continue future changes in a fresh branch (through "rebase and merge", squashing this all will make this commit unreadable).

---

And yes, I'd love to skip merge commits. To remove it, you can do the following:
```bash
git fetch
git checkout cleanup # just to make sure you're on this branch
git rebase origin/main
```
You'll have to handle one conflict (probably the same way you handled the conflcit in the merge commit). Feel free to DM me if you need help or tell me if you'd like me to do it for you :+1:

jfmengels:
I tend to prefer `[ "Basics" ] ->` (even with multiple elements `[ "A", "B", "C" ] ->`) when I know the exact number. So yes, leave them as is

Issue #109, opened 2023-05-10
Add simplifications for elm/random


jfmengels:
```elm
Random.uniform a []
--> Random.constant a

Random.weighted ( weight, a ) []
--> Random.constant a

Random.weighted tuple []
--> Random.constant (Tuple.first tuple)

Random.list 0 generator
--> Random.constant []

Random.map identity generator
--> generator

Random.andThen Random.constant generator
--> generator
```

Also detect when `Random.lazy` is unnecessary?

lue-bird:
Maybe also
```elm
Random.list 0 generator
--> Random.map List.singleton generator

Random.map (always a)
--> Random.constant a
```

jfmengels:
I think I'm on board with
```elm
Random.list 1 generator
--> Random.map List.singleton generator
```
though I could imagine some people finding the former one more readable.

---

Should
```elm
Random.andThen f (Random.constant x)
--> f x
```
be
```elm
Random.andThen f (Random.constant x)
--> Random.constant (f x)
```
? If so this could be done but I could also imagine quickly getting ugly :thinking: 

But let's try it. And I'm on board with the rest :+1:

lue-bird:
No, adding `constant` would be wrong.
Maybe you can make it work in your head if you look at the same simplification for `Maybe`: https://github.com/jfmengels/elm-review-simplify/blob/main/src/Simplify.elm#L372-L373

PR #132, opened 2023-08-16
Add Random.list simplifications

lue-bird:
```elm
Random.list 0 generator
--> Random.constant []

-- not part of the summary
Random.list -1/-2/-3/... generator
--> Random.constant []

Random.list 1 generator
--> Random.map List.singleton generator

Random.list n (Random.constant el)
--> Random.constant (List.repeat n el)
```
part of #109


lue-bird:
The last 2 commits are unrelated.
They only correct the normalization test "should normalize an operator application" with the new `elm-syntax` upgrade so that the CI and the local setup are happy

PR #134, opened 2023-08-17
Add Random.map simplifications

lue-bird:
```elm
Random.map identity generator
--> generator

Random.map (always a) generator
--> Random.constant a

-- not in summary
Random.map << always
--> Random.map

Random.map f (Random.constant x)
--> Random.constant (f x)

-- not in summary
Random.map f << Random.constant
--> Random.constant << f
```
part of https://github.com/jfmengels/elm-review-simplify/issues/109

Some comments
- The generalized checks for `map f (pure x)` and `map identity` are also now re-used for existing `Maybe.map`, `Result.map`, ... checks
    - they are versatile and pretty: you can for example use them with `Ok` and `map` or with `Err` and `mapError`
- As an unrelated change, I upgraded `elm-syntax` to the latest version to make the local tests pass
- I feel like I could do another round of cleanups with better errors for error infos I marked TODO, more performant `compositionChecks` similar to `functionCallChecks,` changing `ModuleName -> String -> ...` to `( ModuleName, String ) -> ...` etc
- The amount of tests here kind of went out of hand (and the list isn't even complete). We could (from now on?) combine cases with equivalent/similar output to get a more focused file, slightly faster and less boring test creation, better performance (not that this has been a problem) but also a bit less clear errors. Thoughts?

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -809,7 +809,7 @@ All of these also apply for `Sub`.
     Random.map identity generator
     --> generator
 
-    Random.map (always a)
+    Random.map (always a) generator
     --> Random.constant a
 
     Random.map f (Random.constant x)
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -6522,6 +6522,22 @@ randomMapChecks checkInfo =
         ()
 
 
+randomMapCompositionChecks : CompositionCheckInfo -> List (Error {})
+randomMapCompositionChecks checkInfo =
+    firstThatReportsError
+        [ \() -> pureToMapCompositionChecks { moduleName = [ "Random" ], pure = "constant", map = "map" } checkInfo
+        , \() -> randomMapAlwaysCompositionChecks checkInfo
+        ]
+        ()
+
+
+randomMapAlwaysErrorInfo : { message : String, details : List String }
+randomMapAlwaysErrorInfo =
+    { message = "Always mapping to the same value is equivalent to Random.constant"
+    , details = [ "Since your Random.map call always produces the same value, you can replace the whole call by Random.constant that value." ]
+    }
+
+
 randomMapAlwaysChecks : CheckInfo -> List (Error {})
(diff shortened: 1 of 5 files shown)


src/Simplify.elm
```diff
@@ -6814,24 +6754,144 @@ collectionMapChecks collection checkInfo =
...
-            else
-                []
         ]
         ()
 
 
+{-| TODO merge with identityError
+-}
+mapIdentityChecks :
+    { a
```
jfmengels, line 6799: We don't an extensible record here.

jfmengels (approved):
Looks good to me :+1: 

And nice refactors for the generalized checks :+1: 

> We could (from now on?) combine cases with equivalent/similar output to get a more focused file, slightly faster and less boring test creation, better performance (not that this has been a problem) but also a bit less clear errors. Thoughts?

Yeah, the tests file is starting to become really big, and too much for my machine to handle :sweat_smile: 
I was thinking we could take the test suites under `all` and split them into different files. But I agree we could have additional files for related concerns, like `map`. I think that makes sense :+1: 

> I feel like I could do another round of cleanups with better errors for error infos I marked TODO, more performant compositionChecks similar to functionCallChecks, changing ModuleName -> String -> ... to ( ModuleName, String ) -> ... etc

If you have ideas, go ahead!

src/Simplify.elm
lue-bird: It's true that we currently don't need it but if you look at for example
```elm
collectionMapChecks collection checkInfo =
    ...
            mapIdentityChecks
                { moduleName = AstHelpers.moduleNameFromString collection.moduleName
                , represents = collection.represents
                }
                checkInfo
```
you can see that the extensibility only accidentally doesn't help because for some reason `Collection` takes `moduleName` as a `String`. How about I make the two compatible in the next cleanup PR?

jfmengels: Sounds good :+1:

PR #135, opened 2023-08-18
Cleanup

lue-bird:
### changes overview
As always, if you dislike how certain changes turned out in the end, say for example [removing isEmptyList in favor of a case of on getListLiteral](https://github.com/jfmengels/elm-review-simplify/pull/135/commits/eefc0c57b5c3f6381c595840f0c72257f1afb450), I'll happily refactor to a consistent style you prefer.
- rework composition
    - introduce a "by later function name" like with functionCallChecks
- always take qualifieds as a tuple, not separate arguments
- prefer `replaceBySubExpressionFix` over `keepOnlyFix` to avoid uncaught parenthesis bugs (and I think this _will_ actually catch a few, like when an if is used as the argument in `<|` but the kept branch is `... |> ...`)
- use `replaceBySubExpressionFix` etc instead of manual fixes to remove complexity and bugs and to avoid unnecessary parens
- always collapse and reduce `getSpecificFunctionCall`, `getSpecificValueOrFunction`
- `Collection` etc get their `moduleName : ModuleName` instead of `moduleName : String`
- prefer `qualifiedToString` instead of raw Strings in error info. This should also make it easier to embed them into future shared error info formats
- refactor away all those `if Maybe.withDefault False (Maybe.map ...` and `case Maybe.andThen ... arg`
- lean into `firstThatReportsError` even more
- correct a bug with `Dict.size (Dict.fromList [b, c])` being replaced by 2 and `Dict.size (Dict.fromList [(1,1), (1,2)]` being replaced by 2 instead of 1
- replace matching on a singleton list literal with `getListSingleton` to get `List.singleton` for free
- rework isAlwaysMaybe, isAlwaysResult, resultValues, maybeValues, ... to `constructs`, `sameInAllBranches`
    - correct bug with e.g. `Result.toMaybe (if x then Err a else Ok b)` being fixed to `Nothing`
    - enable `mapError f (if x then Err a else Err b)` fix to `f (if x then a else b)`
    - correct with e.g. `Maybe.andThen (always (Just a)) maybe` being fixed to `maybe`

### to discuss
- adding
  ```elm
  qualifyToString fullyQualified resources =
      qualifiedToString (qualify fullyQualified resources)`
  ```
    - more convenient and readable
    - easier to mix up with `qualifiedToString`
- changing e.g.
  ```elm
  compositionChecks checkInfo =
      firstThatReportsError
          [ \() -> basicsIdentityCompositionChecks checkInfo
          , \() -> basicsNotCompositionChecks checkInfo
          , \() -> basicsNegateCompositionChecks checkInfo
          , \() -> ...
          ]
          ()
  ```
  to
  ```elm
  compositionChecks =
      firstThatReportsError
          [ basicsIdentityCompositionChecks 
          , basicsNotCompositionChecks 
          , basicsNegateCompositionChecks 
          , \checkInfo -> ...
          ]
  ```
    - the lower one is IMO more readable and the sub-checks are more atomized
    - performance in the lower one might be a tad worse because of currying
- Switching from returning a list of errors for each check to maybe one error
    - seems more in line with what's actually being reported
    - constructing and deconstructing lists might be more expensive than with `Maybe`?
    - `firstThatReportsError` becomes almost `findMap`, which is a bit more intuitive

### follow-up PR
I'd like to to overhaul error infos to common formats (those already marked TODO and many more like adding even more general checks like andThenPureChecks for the types) and polish outliers like `"Unnecessary wildcard argument argument"`.
I could add the changes them here but they probably need much more discussion and generally more time which doesn't help merging the current, unrelated changes in this PR.

I can start this PR work right after this one is merged or later depending on if you want to tweak some things after merge.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1429,17 +1429,18 @@ expressionVisitor node context =
                                 RangeDict.remove expressionRange withNewBranchLocalBindings
                         }
 
-            { errors, rangesToIgnore, rightSidesOfPlusPlus, inferredConstants } =
+            expressionChecked : { errors : List (Error {}), rangesToIgnore : RangeDict (), rightSidesOfPlusPlus : RangeDict (), inferredConstants : List ( Range, Infer.Inferred ) }
+            expressionChecked =
                 expressionVisitorHelp node contextWithInferredConstantsAndLocalBindings
         in
-        ( errors
+        ( expressionChecked.errors
         , { contextWithInferredConstantsAndLocalBindings
-            | rangesToIgnore = RangeDict.union rangesToIgnore context.rangesToIgnore
-            , rightSidesOfPlusPlus = RangeDict.union rightSidesOfPlusPlus context.rightSidesOfPlusPlus
+            | rangesToIgnore = RangeDict.union expressionChecked.rangesToIgnore context.rangesToIgnore
+            , rightSidesOfPlusPlus = RangeDict.union expressionChecked.rightSidesOfPlusPlus context.rightSidesOfPlusPlus
             , inferredConstantsDict =
                 List.foldl (\( range, constants ) acc -> RangeDict.insert range constants acc)
                     contextWithInferredConstants.inferredConstantsDict
-                    inferredConstants
+                    expressionChecked.inferredConstants
           }
         )
 
@@ -3553,9 +3554,14 @@ comparisonError bool checkInfo =
 
 
 targetIfKeyword : Range -> Range
-targetIfKeyword { start } =
-    { start = start
-    , end = { start | column = start.column + 2 }
+targetIfKeyword ifExpressionRange =
+    let
+        ifStart : Location
+        ifStart =
+            ifExpressionRange.start
+    in
+    { start = ifStart
+    , end = { ifStart | column = ifStart.column + 2 }
     }
 
 
@@ -7307,10 +7313,10 @@ pipelineChecks checkInfo =
 
 
 fullyAppliedLambdaInPipelineChecks : { nodeRange : Range, firstArgument : Node Expression, function : Node Expression } -> List (Error {})
-fullyAppliedLambdaInPipelineChecks { nodeRange, function, firstArgument } =
-    case Node.value function of
+fullyAppliedLambdaInPipelineChecks checkInfo =
+    case Node.value checkInfo.function of
         Expression.ParenthesizedExpression (Node lambdaRange (Expression.LambdaExpression lambda)) ->
-            case Node.value (AstHelpers.removeParens firstArgument) of
+            case Node.value (AstHelpers.removeParens checkInfo.firstArgument) of
                 Expression.OperatorApplication "|>" _ _ _ ->
                     []
 
@@ -7319,7 +7325,7 @@ fullyAppliedLambdaInPipelineChecks { nodeRange, function, firstArgument } =
 
                 _ ->
                     appliedLambdaChecks
-                        { nodeRange = nodeRange
+                        { nodeRange = checkInfo.nodeRange
                         , lambdaRange = lambdaRange
                         , lambda = lambda
                         }
@@ -8310,8 +8316,8 @@ fullyAppliedPrefixOperatorChecks checkInfo =
 
 
 appliedLambdaChecks : { nodeRange : Range, lambdaRange : Range, lambda : Expression.Lambda } -> List (Error {})
-appliedLambdaChecks { nodeRange, lambdaRange, lambda } =
-    case lambda.args of
+appliedLambdaChecks checkInfo =
+    case checkInfo.lambda.args of
         (Node unitRange Pattern.UnitPattern) :: otherPatterns ->
             [ Rule.errorWithFix
                 { message = "Unnecessary unit argument"
@@ -8323,11 +8329,11 @@ appliedLambdaChecks { nodeRange, lambdaRange, lambda } =
                 unitRange
                 (case otherPatterns of
                     [] ->
-                        replaceBySubExpressionFix nodeRange lambda.expression
+                        replaceBySubExpressionFix checkInfo.nodeRange checkInfo.lambda.expression
 
                     secondPattern :: _ ->
                         Fix.removeRange { start = unitRange.start, end = (Node.range secondPattern).start }
-                            :: keepOnlyAndParenthesizeFix { parentRange = nodeRange, keep = lambdaRange }
+                            :: keepOnlyAndParenthesizeFix { parentRange = checkInfo.nodeRange, keep = checkInfo.lambdaRange }
                 )
             ]
 
@@ -8342,11 +8348,11 @@ appliedLambdaChecks { nodeRange, lambdaRange, lambda } =
                 allRange
                 (case otherPatterns of
                     [] ->
-                        replaceBySubExpressionFix nodeRange lambda.expression
+                        replaceBySubExpressionFix checkInfo.nodeRange checkInfo.lambda.expression
 
                     secondPattern :: _ ->
                         Fix.removeRange { start = allRange.start, end = (Node.range secondPattern).start }
-                            :: keepOnlyAndParenthesizeFix { parentRange = nodeRange, keep = lambdaRange }
+                            :: keepOnlyAndParenthesizeFix { parentRange = checkInfo.nodeRange, keep = checkInfo.lambdaRange }
                 )
             ]
 
@@ -8362,7 +8368,7 @@ appliedLambdaChecks { nodeRange, lambdaRange, lambda } =
 - Extract the lambda to a named function (at the top-level or defined in a let expression)"""
                     ]
                 }
-                lambdaRange
+                checkInfo.lambdaRange
             ]
 
 
@@ -8881,18 +8887,18 @@ collapsedConsRemoveElementFix :
     , tailRange : Range
     }
     -> List Fix
-collapsedConsRemoveElementFix { toRemove, tailRange } =
-    case ( toRemove.before, toRemove.after ) of
+collapsedConsRemoveElementFix config =
+    case ( config.toRemove.before, config.toRemove.after ) of
         -- found the only consed element
         ( Nothing, Nothing ) ->
             [ Fix.removeRange
-                { start = toRemove.found.range.start, end = tailRange.start }
+                { start = config.toRemove.found.range.start, end = config.tailRange.start }
             ]
 
         -- found first consed element
         ( Nothing, Just (Node afterRange _) ) ->
             [ Fix.removeRange
-                { start = toRemove.found.range.start
+                { start = config.toRemove.found.range.start
                 , end = afterRange.start
                 }
             ]
@@ -8901,7 +8907,7 @@ collapsedConsRemoveElementFix { toRemove, tailRange } =
         ( Just (Node beforeRange _), _ ) ->
             [ Fix.removeRange
                 { start = beforeRange.end
-                , end = toRemove.found.range.end
+                , end = config.toRemove.found.range.end
                 }
             ]
 
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -5575,10 +5575,11 @@ listFilterMapChecks checkInfo =
                                             , details = [ qualifiedToString ( [ "List" ], "filterMap" ) ++ " is meant for this exact purpose and will also be faster." ]
                                             }
                                             { start = checkInfo.fnRange.start, end = (Node.range checkInfo.firstArg).end }
-                                            [ removeFunctionAndFirstArg checkInfo listMapCall.nodeRange
-                                            , Fix.replaceRangeBy listMapCall.fnRange
-                                                (qualifiedToString (qualify ( [ "List" ], "filterMap" ) checkInfo))
-                                            ]
+                                            (replaceBySubExpressionFix checkInfo.parentRange listArg
+                                                ++ [ Fix.replaceRangeBy listMapCall.fnRange
+                                                        (qualifiedToString (qualify ( [ "List" ], "filterMap" ) checkInfo))
+                                                   ]
+                                            )
                                         ]
 
                                     Nothing ->
@@ -8728,15 +8729,6 @@ rangeContainsLocation location =
             ((Range.compareLocations location range.start == LT)
                 || (Range.compareLocations location range.end == GT)
             )
-
-
-removeFunctionAndFirstArg : { a | fnRange : Range, firstArg : Node b, usingRightPizza : Bool } -> Range -> Fix
-removeFunctionAndFirstArg checkInfo secondArgRange =
-    if checkInfo.usingRightPizza then
-        Fix.removeRange { start = secondArgRange.end, end = (Node.range checkInfo.firstArg).end }
-
-    else
-        Fix.removeRange { start = checkInfo.fnRange.start, end = secondArgRange.start }
 
 
 rangeWithoutBoundaries : Range -> Range
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -8328,9 +8328,8 @@ appliedLambdaChecks { nodeRange, lambdaRange, lambda, firstArgument } =
                         replaceBySubExpressionFix nodeRange lambda.expression
 
                     secondPattern :: _ ->
-                        [ Fix.removeRange { start = unitRange.start, end = (Node.range secondPattern).start }
-                        , Fix.removeRange (Node.range firstArgument)
-                        ]
+                        Fix.removeRange { start = unitRange.start, end = (Node.range secondPattern).start }
+                            :: keepOnlyAndParenthesizeFix { parentRange = nodeRange, keep = lambdaRange }
                 )
             ]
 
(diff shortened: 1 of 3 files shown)


tests/SimplifyTest.elm
```diff
@@ -1496,8 +1496,8 @@ a = not >> not
                     |> Review.Test.run ruleWithDefaults
                     |> Review.Test.expectErrors
                         [ Review.Test.error
-                            { message = "Unnecessary double negation"
-                            , details = [ "Composing `not` with `not` cancel each other out." ]
+                            { message = "Unnecessary double Basics.not"
+                            , details = [ "Composing Basics.not with Basics.not cancels each other out." ]
```
jfmengels, line 1500: I don't know what the correct spelling should be here: "cancel" or "cancels"? 🤔

jfmengels:
https://github.com/jfmengels/elm-review-simplify/pull/135/commits/9aed1dc48f7ac82cd1013135f918e1b557049c99 contained a bunch of different changes and was a bit hard to review, it would have been nice if these changes were in different commits.

Here's something I like to do: When I'm doing a change, and I then notice a different thing I should (or want to) change first, I `git stash` my changes so I start again with a clean slate, do my changes, commit, and then `git stash pop` to resume the original work. Sometimes I stack these changes and that's okay (it's actually useful, because it reminds me where I needed to continue work 😄 ). Hope this helps!

jfmengels (commented):
Looks good so far. Feel free to merge the changes you have already 👍

tests/SimplifyTest.elm
lue-bird: Yeah I also don't like either option. The change was mostly because I had to change the error infos anyway. I'll add this to the "to rework list" :+1:

lue-bird:
> When I'm doing a change, and I then notice a different thing I should (or want to) change first, I git stash my changes so I start again with a clean slate, do my changes, commit, and then git stash pop

Oh, that's super helpful!

src/Simplify/AstHelpers.elm
```diff
@@ -71,12 +71,11 @@ removeParensFromPattern patternNode =
 
 isSpecificValueOrFunction : ( ModuleName, String ) -> ModuleNameLookupTable -> Node Expression -> Bool
 isSpecificValueOrFunction ( moduleName, name ) lookupTable expressionNode =
-    case removeParens expressionNode of
-        Node fnRange (Expression.FunctionOrValue _ foundName) ->
-            (foundName == name)
-                && (ModuleNameLookupTable.moduleNameAt lookupTable fnRange == Just moduleName)
+    case getSpecificValueOrFunction ( moduleName, name ) lookupTable expressionNode of
```
jfmengels: For a small perf improvement, we should not destructure `( moduleName, name )` to then recreate such a tuple. That allocates an additional tuple unnecessarily.

(I'm sure we can find more places where we could avoid this problem)

src/Simplify/AstHelpers.elm
```diff
@@ -396,12 +396,12 @@ isIdentity lookupTable baseExpressionNode =
         Expression.FunctionOrValue _ "identity" ->
             ModuleNameLookupTable.moduleNameFor lookupTable expressionWithoutParensNode == Just [ "Basics" ]
 
-        Expression.LambdaExpression { args, expression } ->
-            case args of
+        Expression.LambdaExpression lambda ->
```
jfmengels: Suggestion: Don't have commits named "Refactor X". Instead, say how you changed X. The commit message doesn't have to be great, but refactor is not super helpful/clear.

An additional tip (not to be applied all the time though): if your commit says "Did X and Y", you probably made too many changes at once. That's my way of working at least but to each their own :wink:

src/Simplify/AstHelpers.elm
```diff
@@ -713,30 +713,19 @@ getTuple expressionNode =
...
-                        Nothing
+getBoolPattern lookupTable basePatternNode =
+    case removeParensFromPattern basePatternNode of
+        Node variantPatternRange (Pattern.NamedPattern variantPattern _) ->
+            case ( ModuleNameLookupTable.moduleNameAt lookupTable variantPatternRange, variantPattern.name ) of
+                ( Just [ "Basics" ], "True" ) ->
+                    Just True
+
+                ( Just [ "Basics" ], "False" ) ->
+                    Just False
```
jfmengels: I agree that this is easier to read, but the reason I did it the way it was before was to avoid computing the module name when `variantPattern.name` was not `True` or `False`. Since in practice this function will be rarely called with a node where `variantPattern.name` is one of those two, I think this is worth keeping as it was before for performance reasons.

tests/SimplifyTest.elm
```diff
@@ -16836,7 +16836,7 @@ a = Cmd.map f Cmd.none
                     |> Review.Test.run ruleWithDefaults
                     |> Review.Test.expectErrors
                         [ Review.Test.error
-                            { message = "Using Cmd.map on Cmd.none will result in Cmd.none"
+                            { message = "Using Platform.Cmd.map on Cmd.none will result in Cmd.none"
```
jfmengels, line 16913: I think it's clearer to talk about `Cmd.map` rather than `Platform.Cmd.map`. I don't know how many Elm users know that that is the real name of `Cmd`.

Also for this line, it's inconsistent with `Cmd.none`.

src/Simplify.elm
```diff
@@ -6607,19 +6607,14 @@ collectionMapChecks :
...
-                checkInfo
+        [ \() -> mapIdentityChecks collection checkInfo
         , \() ->
             case secondArg checkInfo of
                 Just collectionArg ->
                     if collection.isEmpty checkInfo.lookupTable collectionArg then
                         [ Rule.errorWithFix
                             -- TODO rework error info
-                            { message = "Using " ++ collection.moduleName ++ ".map on " ++ collection.emptyDescription ++ " will result in " ++ collection.emptyDescription
+                            { message = "Using " ++ qualifiedToString ( collection.moduleName, "map" ) ++ " on " ++ collection.emptyDescription ++ " will result in " ++ collection.emptyDescription
```
jfmengels, line 6815: Not convinced about `qualifiedToString` here either :thinking:  (but happy to be convinced)

src/Simplify.elm
```diff
@@ -4132,32 +4132,37 @@ resultMapErrorChecks checkInfo =
             else
                 []
         , \() ->
-            case Maybe.andThen (AstHelpers.getSpecificFunctionCall ( [ "Result" ], "Err" ) checkInfo.lookupTable) maybeResultArg of
```
jfmengels, line 4204: Personal opinion: Having less indentation and less case branches by using `Maybe.andThen` is actually better (as long as we don't need the intermediate value).

tests/SimplifyTest.elm
```diff
@@ -16483,6 +16484,94 @@ a = Dict.singleton x y |> Dict.size
                             }
                             |> Review.Test.whenFixed """module A exposing (..)
 a = 1
+"""
+                        ]
+        , test "should replace Dict.size (Dict.fromList []) by 0" <|
+            \() ->
+                """module A exposing (..)
+a = Dict.size (Dict.fromList [])
```
jfmengels, line 16492: We're actually missing an import to `Dict` for all of these tests.

tests/SimplifyTest.elm
```diff
@@ -16434,6 +16434,7 @@ dictSizeTests =
 a = Dict.size
 a = Dict.size b
 a = Dict.size (Dict.fromList [b, c, d])
+a = Dict.size (Dict.fromList [((),1), ((),2), ((),3)])
```
jfmengels, line 16437: I'm super confused as to why these last two are not reported. Are they not the same as the tests you've added below?

EDIT: Aaah, gotcha, it's because the number can be wrong if some of the entries have the same key. Maybe edit the title of the new tests to make that explicit, because it was quite confusing to me.
Also, we *should* be simplifying the line I just commented on to `1`. Maybe to be done later, but I don't think it should be part of the "okay" tests.

jfmengels (commented):
I'm at commit https://github.com/jfmengels/elm-review-simplify/pull/135/commits/755d623bc4919db430f2f2bb860cb2f616e0f886, I'll do the rest later :sweat_smile: 

I have a few comments. I think we can push the commits up until here to master still (especially since there id a large number of commits depending on the changes  and they're not deal-breaking).

tests/SimplifyTest.elm
lue-bird: units aren't comparable.
```
> () < ()
-- TYPE MISMATCH ---------------------------------------------------------- REPL

I cannot do a comparison with this value:

3|   () < ()
     ^^
The left side of (<) is a unit value:

    ()

But (<) only works on Int, Float, Char, and String values. It can work on lists
and tuples of comparable values as well, but it is usually better to find a
different path.

Hint: I only know how to compare ints, floats, chars, strings, lists of
comparable values, and tuples of comparable values.
```

src/Simplify.elm
lue-bird: Just to be sure: Do you know the difference between `qualifiedToString` and `qualifiedToString (qualify ...)`?
- just `qualifiedToString` always prints with the exact module name you provide (usually fully qualified), even when for example exposed in the imports
- `qualify` takes a fully qualified thing and adapts the lowest currently allowed qualification. E.g. it sets the module name to `[]` when exposed in the current module

If you did know that and asking why I'm using `qualifiedToString` instead of a raw string, that's just to make it easier in the future to generalize the error and fix format. I'd like to do that in the next PR so this change is in a bit of a limbo

src/Simplify.elm
```diff
@@ -4779,7 +4779,7 @@ dictToListMapChecks : CheckInfo -> List (Error {})
 dictToListMapChecks listMapCheckInfo =
     case secondArg listMapCheckInfo of
         Just listArgument ->
-            case AstHelpers.getSpecificReducedFunctionCall ( [ "Dict" ], "toList" ) listMapCheckInfo.lookupTable listArgument of
+            case AstHelpers.getSpecificFunctionCall ( [ "Dict" ], "toList" ) listMapCheckInfo.lookupTable listArgument of
```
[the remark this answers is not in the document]
lue-bird: We''ll get more stuff simplified yeah.

src/Simplify.elm
```diff
@@ -5218,48 +5285,40 @@ listFoldAnyDirectionChecks foldOperationName checkInfo =
...
+                        )
+                    ]
 
-                -- reduce argument is missing
-                [] ->
+                _ ->
                     []
 
-        Nothing ->
+        -- composition onto fully constructed value (compile-time error)
```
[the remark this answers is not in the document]
lue-bird: true

src/Simplify.elm
```diff
@@ -6310,9 +6321,18 @@ randomListChecks checkInfo =
...
                     case AstHelpers.getSpecificFunctionCall ( [ "Random" ], "constant" ) checkInfo.lookupTable elementGeneratorArg of
                         Just constantCall ->
+                            let
+                                currentAsString : String
+                                currentAsString =
+                                    qualifiedToString ( [ "Random" ], "list" ) ++ " n (" ++ qualifiedToString ( [ "Random" ], "constant" ) ++ " el)"
+
+                                replacementAsString : String
+                                replacementAsString =
+                                    qualifiedToString ( [ "Random" ], "constant" ) ++ " (List.repeat n el)"
```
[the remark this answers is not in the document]
lue-bird: Already done in a later commit :sweat_smile:

src/Simplify.elm
```diff
@@ -3667,7 +3667,7 @@ stringFromListChecks checkInfo =
     case Node.value checkInfo.firstArg of
         Expression.ListExpr [] ->
             [ Rule.errorWithFix
-                { message = "Calling String.fromList [] will result in " ++ emptyStringAsString
+                { message = "Calling " ++ qualifiedToString ( [ "String" ], "fromList" ) ++ " [] will result in " ++ emptyStringAsString
```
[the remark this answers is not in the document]
lue-bird: I think using `qualify` would be a mistake and I don't plan on proposing that as a change. Made these changes mostly to make changing to a more general format easier, see my comment further below https://github.com/jfmengels/elm-review-simplify/pull/135#discussion_r1302217518

src/Simplify.elm
lue-bird: How about a rule like "if it fits on one line it's fine to keep the `case andThen of`?"

src/Simplify/AstHelpers.elm
lue-bird: Ok, reverted.

This to me strongly smells to me like something a smart compiler could automatically optimize (checking already evaluated things first).

I guess elm's really strong eager eval always constructs the tuple, even if we reversed the order so that option also doesn't exist sadg

src/Simplify.elm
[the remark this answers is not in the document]
jfmengels: Nice :+1:

src/Simplify/AstHelpers.elm
jfmengels: Yeah, this is a benefit of laziness that we don't have. Not sure if it would apply in this specific case though :thinking:

tests/SimplifyTest.elm
jfmengels: Oh yeah they aren't. But that also means that technically `Dict.fromList [ ( (), 1 ) ]` doesn't compile :smile:

lue-bird: very true as well, removed those from the tests

src/Simplify.elm
```diff
@@ -8872,89 +8872,70 @@ needsParens expr =
...
-        Expression.FunctionOrValue _ "Just" ->
-            case ModuleNameLookupTable.moduleNameFor lookupTable expressionWithoutParensNode of
-                Just [ "Maybe" ] ->
-                    Determined
-                        (Just
-                            { fixes = [ Fix.removeRange (Node.range expressionWithoutParensNode) ]
-                            , throughLambdaFunction = False
-                            }
-                        )
+    case AstHelpers.getSpecificValueOrFunction ( [ "Result" ], "Err" ) lookupTable baseExpressionNode of
```
jfmengels: Comment on commit description: If I see a commit saying "Rename X to Y", I will tend to believe that you *only* did that change. And because in Elm renaming things is very safe, I will in practice skip reviewing that commit. I'll at most look at the old name and the new name. I'll try to use the same logic when I need to understand in which commit a bug was introduced to find the commit faster.

Here, it looks like you did other things though, which is confusing. I highly recommend that when a task is splittable, that you do it in 2 commits.

Another example of where splitting things in 2 commits would make reviewing a commit much easier is when you introduce `firstThatReportsError` in a function. Refactoring a function to use it is pretty simple and predictable, to the point where if you have a commit saying "Make X use firstThatReportsError" I would skim over it. And then have a separate commit to introduce different changes like using a different AstHelpers function. Combining two two, with the added indentation makes it really much harder to see what has or has not changed.

On the topic, I really enjoyed [this talk](https://www.youtube.com/watch?v=wp6oSVDdbXQ&pp=ygUPbGxld2VsbHluIGZhbGNv) :relaxed:

src/Simplify/AstHelpers.elm
```diff
@@ -861,11 +861,12 @@ If desired, call in combination with `qualify`
...
-    if List.isEmpty moduleName then
-        name
+    case moduleName of
+        [] ->
+            name
 
-    else
-        moduleNameToString moduleName ++ "." ++ name
+        moduleNamePart0 :: moduleNamePart1Up ->
+            moduleNameToString (moduleNamePart0 :: moduleNamePart1Up) ++ "." ++ name
```
jfmengels, line 923: Not fond of changes like these where we deconstruct lists to reconstruct them right after, mostly for performance reasons :thinking:

lue-bird: Ok, used the variable we cased on instead of reconstructing. Add a comment if you find more cases like that.

(All these again to me seem optimizable by some tool. Kinda sad that we have to make these micro-optimizations here.)

jfmengels: (not going to disagree with you here :/ )

jfmengels (approved):
Let's merge it!

jfmengels:
Thank you for this extremely large and great changes! The commit story was much better in the second half, much more enjoyable to review :relaxed: 

I hope you had fun doing these changes!

lue-bird:
Thanks for reviewing! Did you take a look at the "to discuss" section in the PR summary?

jfmengels:
> Did you take a look at the "to discuss" section in the PR summary?

Oops, forgot to reply to those.

`qualifyToString` sounds good to me :+1: 

Let's do your proposed changes for `compositionChecks`, I was thinking of that as well

> Switching from returning a list of errors for each check to maybe one error

That sounds good to me. I guess there are very few places (if any) where we return multiple errors, right?

> I can start this PR work right after this one is merged or later depending on if you want to tweak some things after merge.

Feel free to do it, I don't plan on tweaking the code at the moment.

jfmengels:
Oh, you mentioned you fixed a few bugs. We should probably mention that in the CHANGELOG.md as well, and ideally write some tests if you didn't already.

PR #146, opened 2023-09-15
Update changelog

lue-bird:
Date is set to 2023-09-16, so if you want to release today (15th) or after tomorrow, we have to change it.

(Unreleted note: Today I learned why I had so much pain with `git` before... `git fetch` doesn't do anything really, only `git fetch --all`.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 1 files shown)


jfmengels:
Thanks! I'll see when I'll release this. I might do this after testing the changes on my project at work on monday, but I'll take care of the date changes and formatting it up like the rest.

Issue #149, opened 2023-09-16
Apply simplifications for ++ when encountering List.singleton


jfmengels:
Currently, the following simplification is applied:
```elm
[ 1 ] ++ [ 2 ]
--> [ 1, 2 ]
```
and there are more and more checks where we consider `[ 1 ]` the same as `List.singleton 1`. But we don't support the following simplification:
```elm
List.singleton 1 ++ [ 2 ]
--> [ 1, 2 ]

[ 1 ] ++ List.singleton 2
--> [ 1, 2 ]
```
I guess we could also do
```elm
List.singleton 1 ++ List.singleton 2
--> [ 1, 2 ]
```
but I could understand some push-back on that.

jfmengels:
Nevermind, I was comparing with older versions and got confused. All of my examples get fixed.
For instance, `List.singleton 1 ++ List.singleton 2` gets fixed to `1 :: List.singleton 2`.

Should we push it further and fix `1 :: List.singleton 2` to `[ 1, 2 ]` though?

PR #151, opened 2023-09-16
Add `a :: List.singleton b --> [ a, b ]` simplification

lue-bird:
Implements #149 + test.

`a :: List.singleton b --> [ a, b ]` is not added to the summary because `a :: [ b ] --> [ a, b ]` already exists there.


jfmengels (approved):
This is a really nice simplification. Not sure how often it will trigger, but it's a nice one :smile:

Issue #133, opened 2023-08-17
Add `List.repeat 1 --> List.singleton`


lue-bird:
Just like `Random.list 1 --> Random.map List.singleton` from #109 
```elm
List.repeat 1 --> List.singleton
```

PR #152, opened 2023-09-16
Add `List.repeat 1 --> List.singleton` simplification

lue-bird:
Implements #133 + tests:
```elm
List.repeat 1 x
--> List.singleton x
```
Interestingly, there was previously a test that _explicitly_ required `List.repeat 1 x` to not be simplified. Do you still know about why you added it?

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -621,6 +621,9 @@ Destructuring using case expressions
     List.repeat 0 x
     --> []
 
+    List.repeat 1 x
+    --> List.singleton x
+
     List.partition f []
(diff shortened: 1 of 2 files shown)


jfmengels:
Yeah I think I know why, and it's not for a good reason.
Seems like at some point I fixed `List.repeat 1 list` to `list` when it should have been `[ list ]`. I noticed the error a bit later and fixed it (in 489b5c6db256f43442f3a6c2f6ce90d003747156) bit did not think about fixing it to use `List.singleton` or wrapping it in `[]`.

PR #159, opened 2023-09-17
Add Random.andThen simplifications

lue-bird:
With that, #109 is fully implemented.
```elm
Random.andThen f (Random.constant x)
--> f x

Random.andThen Random.constant generator
--> generator

Random.andThen (\a -> Random.constant b) generator
--> Random.map (\a -> b) generator

Random.andThen (always thenGenerator) generator
--> thenGenerator

-- not in summary
Random.andThen (always thenGenerator)
--> (always thenGenerator)
```

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -851,6 +851,18 @@ All of these also apply for `Sub`.
     Random.map f (Random.constant x)
     --> Random.constant (f x)
 
+    Random.andThen f (Random.constant x)
+    --> f x
+
+    Random.andThen Random.constant generator
+    --> generator
+
+    Random.andThen (\a -> Random.constant b) generator
+    --> Random.map (\a -> b) generator
+
+    Random.andThen (always generator)
+    --> generator
+
 -}
 
 import Dict exposing (Dict)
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -6105,6 +6118,51 @@ randomMapCompositionChecks checkInfo =
...
+        Just alwaysResult ->
+            Just
+                (let
+                    replacementAndFix : { replacementDescription : String, fix : List Fix }
+                    replacementAndFix =
+                        case secondArg checkInfo of
+                            Nothing ->
+                                { replacementDescription = "always with the " ++ wrapper.represents ++ " produced by the function"
+                                , fix =
+                                    -- maybe a bit too clever?
```
jfmengels, line 6145: Why do you think it's too clever?
I'm not 100% confident but I think this is fine (at least with the current way `getAlwaysResult` is implemented.

lue-bird: Yeah no it should be fine in practice. I just thought that conceptually you aren't really "returning the given function" but "returning always with the value returned by the given function". They both just accidentally align which is why I added the comment :)

jfmengels: Yeah, remove that comment :+1:

PR #163, opened 2023-09-17
Add Tuple.pair a b --> ( a, b )

lue-bird:
To complete the suggested tuple simplifications from #2
```elm
Tuple.pair a b
--> ( a, b )
```
It's pretty easy to mess indentation up with these fixes that use `extractSourceCode`, so the fix preserves it 1:1.

I'm not really sure about the error details on this one. Why is the code obviously better after the simplification?


jfmengels:
This also changes
```elm
a
  |> f
  |> Tuple.pair g
-->
( g,
  a
  |> f
)
```

which I don't think we want in this specific instance. Pipelining (and partial application) is almost the only "valid" uses I see/have for `Tuple.pair`.

PR #164, opened 2023-09-18
Add Task.andThen simplifications

lue-bird:
`Task.andThen` simplifications as hinted at in #2 and more.
These are equivalent to `Result.andThen` with different type properties so effectively free.
```elm
Task.andThen f (Task.fail x)
--> Task.fail x
Task.andThen f (Task.succeed a)
--> f a
Task.andThen Task.succeed task
--> task
Task.andThen (\a -> Task.succeed b) task
--> Task.map (\a -> b) x
```
`map` and `mapN` are also effectively free. Same with `onError` (andThen with fail as wrap) and `mapError` (map with fail as wrap).

### DIFF
--- a/tests/SimplifyTest.elm
+++ b/tests/SimplifyTest.elm
@@ -33,6 +33,7 @@ all =
         , dictSimplificationTests
         , cmdTests
         , subTests
+        , taskTests
         , parserTests
         , jsonDecodeTests
         , htmlAttributesTests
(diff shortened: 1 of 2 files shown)


tests/SimplifyTest.elm
```diff
@@ -18256,6 +18257,156 @@ a = Sub.none
...
+"""
+                    |> Review.Test.run ruleWithDefaults
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Task.andThen on a failing task will result in the given failing task"
+                            , details = [ "You can replace this call by the given failing task." ]
+                            , under = "Task.andThen"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = (Task.fail z)
```
jfmengels, line 18293: Could this lead to compiler errors because the type annotation is wrong? I don't think it could, but double-checking.

jfmengels (approved):
This looks good to me. The only thing I'm worried about is making sure that we don't discard tasks. But I think that's more of an issue with `Task.map` than `Task.andThen`. Do you concur?

tests/SimplifyTest.elm
lue-bird: andThen can't change the error type and `fail` produced a free variable for the success value (just like with result). So we should be good!

jfmengels:
Nice, removing these from #2 :relaxed:

PR #166, opened 2023-09-18
Add Task.sequence simplifications

lue-bird:
Implementation should also make it easy to extend to `Maybe/Result/Random/Parser.Extra.sequence/combine` in the future
```elm
Task.sequence [ Task.succeed a, Task.succeed b ]
--> Task.succeed [ a, b ]

Task.sequence [ Task.succeed a, Task.fail x ]
--> Task.fail x

Task.sequence [ task ]
--> Task.map List.singleton task

Task.sequence [ a, Task.fail x, b ]
--> Task.sequence [ a, Task.fail x ]

-- for example not in summary
Task.sequence << List.singleton
--> Task.map List.singleton

-- not in summary because theoretically same case as the first but with a separate error
Task.sequence []
--> Task.succeed []
```
Note that we do not simplify
```elm
Task.sequence [ task, Task.fail x ]
--> Task.fail x
```
because `task` could fail earlier.
We also don't simplify it to
```elm
Task.sequence [ task, Task.fail x ]
--> Task.onError (always (Task.fail x)) task
```
because this is arguably a code style choice.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -845,6 +845,9 @@ All of these also apply for `Sub`.
     Task.sequence [ Task.succeed a, Task.fail x ]
     --> Task.fail x
 
+    Task.sequence [ a, Task.fail x, b ]
+    --> Task.sequence [ a, Task.fail x ]
+
     Task.sequence [ task ]
     --> Task.map List.singleton task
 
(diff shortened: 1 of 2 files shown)


lue-bird:
Setting to draft because I will also add
```elm
Task.sequence [ a, Task.fail x, b ]
--> Task.sequence [ a, Task.fail ]
```

src/Simplify.elm
```diff
@@ -6025,6 +6036,156 @@ taskOnErrorChecks checkInfo =
...
+                checkInfo
+        , \() ->
+            case AstHelpers.getListSingleton checkInfo.lookupTable checkInfo.firstArg of
+                Just singletonList ->
+                    let
+                        mapFn : ( ModuleName, String )
+                        mapFn =
+                            ( wrapper.moduleName, wrapper.mapFnName )
+
+                        replacement qualifyResources =
```
jfmengels, line 6146: This doesn't have to be a function, right? This could already be a constant that uses `checkInfo` instead of `qualifyResources`.

jfmengels (approved):

src/Simplify.elm
lue-bird: It's used with `defaultQualifyResources` in the error info and with `checkInfo` in the fix.
Adding extra qualification in the error is nice I think? This is what we do for all other errors at least.

jfmengels: What I mean is that there is no need to have `replacement` be a function in this context of this function. You could have written.

```elm
replacement =
      qualifiedToString (qualify mapFn checkInfo)
          ++ " "
          ++ qualifiedToString (qualify ( [ "List" ], "singleton" ) checkInfo)
```

I see the value of making this a function if it was possible that we wouldn't need to compute the replacement (based on conditions evaluated later), but in this case there is no more branching that would make this function not be called with `checkInfo`.

lue-bird: I think we're both misunderstanding...
The let declared function _is used with different arguments_ in different contexts here.
For the fix, it's always called with `checkInfo`, just like you've pointed out.
But in the error message and error details, you'll find calls to `replacement defaultQualifyResources` which for example return `Task.sequence ...` instead of `sequence ...` even when `sequence` was imported and exposed in the relevant module.

jfmengels: Aha! Alright, that makes sense! Sorry and thank you for your patience :sweat_smile:

PR #167, opened 2023-09-19
Add Task.map simplifications

lue-bird:
```elm
Task.map identity task
--> task

Task.map f (Task.fail x)
--> Task.fail x

Task.map f (Task.succeed a)
--> Task.succeed (f a)

-- not included in summary for example
Task.map f << Task.succeed
--> Task.succeed << f
```

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -815,6 +815,15 @@ All of these also apply for `Sub`.
 
 ### Task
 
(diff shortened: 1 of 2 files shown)


jfmengels (approved):
Excellent work as always. I'm sure you're pretty glad with how easy all of these are to add now :smile:

PR #169, opened 2023-09-20
Add Array.map simplifications

jfmengels:


src/Simplify.elm
```diff
@@ -6981,6 +6991,64 @@ stringDetermineLength expression =
...
+                Just fromListCall ->
+                    case AstHelpers.getListLiteral fromListCall.firstArg of
+                        Just [] ->
+                            Just (Exactly 0)
+
+                        Just (_ :: []) ->
+                            Just (Exactly 1)
+
+                        Just (el0 :: el1 :: el2Up) ->
+                            case traverse getComparableExpression (el0 :: el1 :: el2Up) of
```
lue-bird, line 7036: I think you copied this from set or somewhere. comparable etc certainly doesn't apply here.
I'd just use `listDetermineLength` for the list from `Array.fromList`

lue-bird (changes requested):

tests/SimplifyTest.elm
```diff
@@ -13944,6 +13945,193 @@ a = identity
...
+                            { message = "Array.map with an identity function will always return the same given array"
+                            , details = [ "You can replace this call by the array itself." ]
+                            , under = "Array.map"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+import Array
+a = x
+"""
+                        ]
+        , test "should replace x |> Array.map identity by x" <|
```
lue-bird, line 14059: For all those `x` variables I'd prefer `array` but it doesn't matter much

jfmengels: (Let me know if you'd prefer me avoiding to force-push in the future)

src/Simplify.elm
jfmengels: :+1:

lue-bird (approved):

PR #170, opened 2023-09-20
Improve variable names in tests

jfmengels:
Also includes a fix to a test title which was incorrect.


jfmengels:
There are probably a lot more remaining though, but this is a good start.

PR #173, opened 2023-09-20
Add wrapper mapN simplifications

lue-bird:
Since we don't want unused code around, this is already applied to `Result.map2-5`:
```elm
-- the following simplifications for map3 work for all Result.mapN
Result.map3 f (Ok a) (Ok b) (Ok c)
--> Ok (f a b c)

Result.map3 f (Ok a) (Err x) thirdResult
--> Err x

Result.map3 f firstResult (Err x) thirdResult
--> Result.map2 f firstResult (Err x)
```
includes all combinations with currying, for example
```elm
Result.map3 f firstResult (Err x)
--> always (Result.map2 f firstResult (Err x))

Result.map4 f firstResult (Err x)
--> (\_ _ -> Result.map2 f firstResult (Err x))
```
The same checks can be applied to `Task`, `Json.Decode`, `Fuzz` etc. with minimal effort.

Note that we don't simplify for example
```elm
Result.map2 (\a b -> a + b) firtResult (Ok second)
--> Result.map (\a -> a + second) firstResult
```
because some consider this a valid code style, e.g.
```elm
Result.map2 YourRecord
    (Json.Decode.field "a" aDecoder)
    (Json.Decode.succeed second)
    (Json.Decode.field "c" cDecoder)
```
I personally disagree strongly but at the same time I'd prefer this to be a separate rule (the fixes are also not trivial, not sure we want to alter the existing code this much).

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -410,6 +410,16 @@ Destructuring using case expressions
     Result.map f (Ok x)
     --> Ok (f x)
 
+    -- the following simplifications for map3 work for all Result.mapN
+    Result.map3 f (Ok a) (Ok b) (Ok c)
+    --> Ok (f a b c)
+
+    Result.map3 f (Ok a) (Err x) thirdResult
+    --> Err x
+
+    Result.map3 f firstResult (Err x) thirdResult
+    --> Result.map2 f firstResult (Err x)
+
     Result.mapError identity x
     --> x
 
@@ -2357,6 +2367,10 @@ functionCallChecks =
         , ( ( [ "Maybe" ], "andThen" ), maybeAndThenChecks )
         , ( ( [ "Maybe" ], "withDefault" ), withDefaultChecks maybeWithJustAsWrap )
         , ( ( [ "Result" ], "map" ), resultMapChecks )
+        , ( ( [ "Result" ], "map2" ), resultMapNChecks { n = 2 } )
+        , ( ( [ "Result" ], "map3" ), resultMapNChecks { n = 3 } )
+        , ( ( [ "Result" ], "map4" ), resultMapNChecks { n = 4 } )
+        , ( ( [ "Result" ], "map5" ), resultMapNChecks { n = 5 } )
         , ( ( [ "Result" ], "mapError" ), resultMapErrorChecks )
         , ( ( [ "Result" ], "andThen" ), resultAndThenChecks )
         , ( ( [ "Result" ], "withDefault" ), withDefaultChecks resultWithOkAsWrap )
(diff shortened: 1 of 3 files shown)


src/Simplify.elm
```diff
@@ -6009,6 +6032,197 @@ emptiableMapNChecks { n } emptiable checkInfo =
         Nothing
 
 
+wrapperMapNChecks : { n : Int } -> WrapperProperties otherProperties -> CheckInfo -> Maybe (Error {})
```
jfmengels, line 6035: I think it would be good to have some documentation (here and on similar functions) that explains what the check does. Because these have become quite generic, even by reading the code I find it pretty hard to figure out what the check does. In practice, the best way for me to find this out is by changing the code so that this function does not do anything anymore, and then look at which tests fail.

I'm thinking we could do something like this (off the top of my head, feel f

```elm
{-| Checks for `mapN` functions that where one of the wrappers causes other arguments to be discarded.

Here are the kinds of simplifications it looks for (examples using Result):

	Result.map2 f (Err y) (Ok x)
        --> Err y

        -- <and then more examples>
-}
wrapperMapNChecks : -- ...
```

What do you think? (I'm fine with adding these later and over time)

jfmengels: Case in point, I'm now realizing that the simplification I added to the example is for the other check function you've introduced :sweat_smile:

src/Simplify.elm
```diff
@@ -6009,6 +6032,197 @@ emptiableMapNChecks { n } emptiable checkInfo =
...
+                                        replaceBySubExpressionFix checkInfo.parentRange emptyAndBefore.found
+                                            ++ [ Fix.insertAt checkInfo.parentRange.start (qualifiedToString (qualify ( [ "Basics" ], "always" ) checkInfo) ++ " ") ]
+                                    }
+
+                                -- multiple args curried
+                                atLeast2 ->
+                                    let
+                                        lambdaStart : String
+                                        lambdaStart =
+                                            "\\" ++ String.join " " (List.repeat atLeast2 "_") ++ " -> "
```
jfmengels, line 6117:
```suggestion
                                            "\\" ++ String.repeat atLeast2 "_ " ++ "-> "
```
(*almost* looks like a new simplification we could look into having?)

src/Simplify.elm
```diff
@@ -6032,6 +6032,20 @@ emptiableMapNChecks { n } emptiable checkInfo =
...
+
+    map2 f (wrap first) (wrap second)
+    --> wrap (f first second)
+
+For example given `resultWithOkAsWrap`:
+
+    Result.map2 f (Ok first) (Ok second)
+    --> Ok (f first second)
+
+This is pretty similar to `wrapperSequenceChecks` where we look at arguments instead of list elements.
```
jfmengels, line 6046: This is super clear, much nicer!

tests/SimplifyTest.elm
```diff
@@ -15406,10 +15406,43 @@ c = Result.map3 f result0
...
+"""
+                    |> Review.Test.run ruleWithDefaults
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Result.map3 where each result is an okay result will result in Ok on the values inside"
+                            , details = [ "You can replace this call by Ok with the function applied to the values inside each okay result." ]
+                            , under = "Result.map3"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = Ok ((c |> g) |> f a b)
```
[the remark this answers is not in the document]
lue-bird: Ok, changed.
`usingRightPizza` doesn't let us check for `<|` so I just used application when False.

I think we should switch to `FunctionCallStyle = Application | Pipe LeftOrRightDirection` in the future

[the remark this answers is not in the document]
jfmengels: Yeah, let's do it. I just tested it with `<|` and I confirm we still have the same problem there, though it's less of an issue because multiline `<|` piping is very rare in Elm.

PR #178, opened 2023-09-21
Add Array.isEmpty simplifications

jfmengels:
Adds all the simplifications for `Array.isEmpty` mention in #169 

- [x] `Array.isEmpty (Array.fromList [x])` -> `False`
- [x] `Array.isEmpty (Array.fromList [])` -> `True`
- [x] `Array.isEmpty Array.empty` -> `True`
- [x] `Array.isEmpty (Array.repeat 2 x)` -> `False`
- [x] `Array.isEmpty (Array.initialize 2 f)` -> `False`

I didn't mention the `Array.isEmpty (Array.initialize 2 f)` to `False` (same for repeat) in the changelog nor documentation.

At the moment, we get 2 errors when doing `Array.isEmpty (Array.fromList [])`. What do you think, is that a problem? I think it's *okay*, and I don't have a nice way around this. We'll have the same problem if we see `repeat 0` or `repeat 1` (or with `initialize`).

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -5771,7 +5771,7 @@ listRepeatChecks checkInfo =
         ()
 
 
-emptiableRepeatChecks : CollectionProperties a otherProperties -> CheckInfo -> Maybe (Error {})
+emptiableRepeatChecks : CollectionProperties otherProperties -> CheckInfo -> Maybe (Error {})
 emptiableRepeatChecks collection checkInfo =
     case Evaluate.getInt checkInfo checkInfo.firstArg of
         Just intValue ->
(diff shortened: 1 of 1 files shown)


src/Simplify.elm
```diff
@@ -6710,11 +6714,11 @@ type alias ConstructWithOneArgProperties =
 
 {-| Properties of a type with with multiple elements. Includes `EmptiableProperties`.
 -}
-type alias CollectionProperties otherProperties =
```
lue-bird, line 6713: There's a much better option:
Taking `Infer.Resources {}` and having a separate `extractInferResources : Infer.Resources a -> Infer.Resources {}`.

You can see how this is done by looking at `asString` which takes `QualifyResources {}` instead of `QualifyResources a`.

lue-bird (changes requested):

src/Simplify.elm
jfmengels: Nice, I'll try that :+1:

lue-bird (approved):

Issue #174, opened 2023-09-21
Simplifications for Array


jfmengels:
Seems like we have no simplifications for any `Array` functions. Don't know how I missed that because there's no work nor issues on that :sweat_smile: 

- [x] [`Array.map`](https://package.elm-lang.org/packages/elm/core/latest/Array#map) #169
- [x] [`Array.filter`](https://package.elm-lang.org/packages/elm/core/latest/Array#filter) #171
- [ ] [`Array.indexedMap`](https://package.elm-lang.org/packages/elm/core/latest/Array#indexedMap)
	- [ ] Same as for `List.indexedMap`?
- [ ] [`Array.repeat`](https://package.elm-lang.org/packages/elm/core/latest/Array#repeat)
  - [ ] `Array.repeat 0 x` -> `Array.empty`
- [ ] [`Array.initialize`](https://package.elm-lang.org/packages/elm/core/latest/Array#initialize)
  - [ ] `Array.initialize 0 f` -> `Array.empty`
- [ ] [`Array.length`](https://package.elm-lang.org/packages/elm/core/latest/Array#length)
  - [ ] `Array.length Array.empty` -> `0`
  - [ ] `Array.length (Array.fromList [ a, b, c ])` -> `3`
  - [ ] `Array.length (Array.repeat 3 x)` -> `3`
  - [ ] `Array.length (Array.initialize 3 f)` -> `3`
- [ ] [`Array.isEmpty`](https://package.elm-lang.org/packages/elm/core/latest/Array#isEmpty)
  - [ ] `Array.isEmpty (Array.fromList [x])` -> `False`
  - [ ] `Array.isEmpty (Array.fromList [])` -> `True`
  - [ ] `Array.isEmpty Array.empty` -> `True`
  - [ ] `Array.isEmpty (Array.repeat 1 x)` -> `False`
  - [ ] `Array.isEmpty (Array.initialize 1 f)` -> `False`
- [ ] [`Array.get`](https://package.elm-lang.org/packages/elm/core/latest/Array#get)
  - [ ] `Array.get n Array.empty` -> `Nothing`
  - [ ] `Array.get -3 array` -> `Nothing`
  - [ ] `Array.get 2 (Array.fromList [ 1, 2, 3 ])` -> `Just 3`
  - [ ] `Array.get 200 (Array.fromList [ 1, 2, 3 ])` -> `Nothing`
  - [ ] `Array.get 2 (Array.repeat 10 x)` -> `Just x`
[...]

PR #179, opened 2023-09-21
Add Array.repeat simplifications

jfmengels:
Adds simplifications for `Array.repeat` #174

- [x] `Array.repeat 0 x` -> `Array.empty`
- [x] `Array.repeat -1 x` -> `Array.empty`

I also wanted to do
```elm
Array.repeat 1 f -- Only when the number is 1
--> Array.fromList [ x ]
```
but the existing checks (`wrapperRepeatChecks`) don't work for this type which doesn't have a `Array.singleton` function.
@lue-bird if you can think of an easy way to add this, let me know, otherwise I'll leave it for later and focus on other issues first :smile:

### DIFF
--- a/tests/SimplifyTest.elm
+++ b/tests/SimplifyTest.elm
@@ -13987,6 +13987,7 @@ arrayTests =
         , arrayMapTests
         , arrayFilterTests
         , arrayIsEmptyTests
+        , arrayRepeatTests
         ]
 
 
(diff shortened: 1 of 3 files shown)


lue-bird:
Yeah, there's no smart mechanism to allow this. I'd just add a custom check.

jfmengels:
I've been trying to simplify `List.repeat 1` but I'm not sure about the simplifications outside of the simple `List.repeat 1 x` case.

For instance, the following solution doesn't feel like an improvement (It's surprising how odd it feels to not have a `Array.singleton` function...) :
```elm
List.repeat 1
--> List.singleton >> Array.fromList
```

Not sure what to do when seeing it in a pipeline either, like
```elm
x 
  |> f
  |> Array.repeat 1
  |> g
-->
[ x 
  |> f
]
  |> Array.fromList
  |> g
```

I'm thinking we should not replace `List.repeat 1`, and maybe only do the simple `List.repeat 1 x` and `List.repeat 1 <| x` cases.

tests/SimplifyTest.elm
```diff
@@ -14617,6 +14618,85 @@ a = Array.isEmpty (Array.initialize n f)
...
+                            |> Review.Test.whenFixed """module A exposing (..)
+import Array
+a = always Array.empty
+"""
+                        ]
+        , test "should replace Array.repeat -5 x by Array.empty" <|
+            \() ->
+                """module A exposing (..)
+import Array
+a = Array.repeat -5 x
```
lue-bird, line 14683: _Maybe_ say `array` instead of `x` (also in the "not reported" and "repeat 0" tests)

lue-bird (approved):

lue-bird:
And yeah those cases all seem very ugly. Might just be best to always leave `Array.repeat 1` as is.

tests/SimplifyTest.elm
jfmengels: That's on purpose, since these are elements and not arrays (well, unless we want `Array (Array Something)`)

lue-bird: oops, yeah true.

PR #180, opened 2023-09-21
Add Task.mapN simplifications

lue-bird:
Just like #173 
```elm
-- the following simplifications for map3 work for all Task.mapN
Task.map3 f (Task.succeed a) (Task.succeed b) (Task.succeed c)
--> Task.succeed (f a b c)

Task.map3 f (Task.succeed a) (Task.fail x) thirdTask
--> Task.fail x

Task.map3 f firstTask (Task.fail x) thirdTask
--> Task.map2 f firstTask (Task.fail x)
```

### DIFF
--- a/tests/SimplifyTest.elm
+++ b/tests/SimplifyTest.elm
@@ -19118,6 +19118,7 @@ taskTests : Test
 taskTests =
     Test.describe "Task"
         [ taskMapTests
+        , taskMapNTests
         , taskAndThenTests
         , taskMapErrorTests
         , taskOnErrorTests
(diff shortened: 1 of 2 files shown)


jfmengels:
I was really unsure about removing tasks but TIL

> Say we were doing HTTP requests instead. `map2` does each task in order, so it would try the first request and only continue after it succeeds. If it fails, the whole thing fails!

tests/SimplifyTest.elm
```diff
@@ -19373,6 +19374,208 @@ a = Task.succeed << f
...
+                            |> Review.Test.whenFixed """module A exposing (..)
+import Task
+a = (Task.fail x)
+"""
+                        ]
+        , test "should replace Task.map3 f (Task.succeed a) (Task.fail x) by always (Task.fail x)" <|
+            \() ->
+                """module A exposing (..)
+import Task
+a = Task.map3 f (Task.succeed a) (Task.fail x)
```
jfmengels, line 19454: We should probably stop using `a` in the references, seeing how that conflicts with the `a =` declaration (making this code that would not compile). Let's try to avoid this in the future and maybe fix this over time, but not necessary for this PR.

lue-bird: right, there are so many places with this issue.

tests/SimplifyTest.elm
```diff
@@ -19373,6 +19374,208 @@ a = Task.succeed << f
...
+            \() ->
+                """module A exposing (..)
+import Task
+a = Task.map3 f (Task.fail x) task1 task2
+"""
+                    |> Review.Test.run ruleWithDefaults
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Task.map3 where we know the first failing task will result in that failing task"
+                            , details = [ "You can replace this call by the first failing task." ]
```
jfmengels, line 19478: I just noticed that the wording is different if the first element is failing than if it's the second one, and I'm not sure why the first argument is special here :thinking:

jfmengels (approved):
Looks good to me. I'm happy that this is possible, as I didn't think it was possible :relaxed:

tests/SimplifyTest.elm
lue-bird: Because they are different situations: One is for when we only know that arguments after are irrelevant because tasks earlier could fail, too. The other replaces the whole expression by the first _known_ failing task.

Or am I missing something?

jfmengels: No that clears it up, thanks 👍

PR #181, opened 2023-09-21
Respect function call style

lue-bird:
Replaces `usingRightPizza : Bool` by `callStyle : FunctionCallStyle` where
```elm
FunctionCallStyle = Application | Pipe LeftOrRightDirection
```

wrapper mapN and map wrap now get fixes that preserve the parent's call style.


jfmengels (approved):
Really nice ❤️

PR #182, opened 2023-09-22
Add String.fromList-toList simplifications

lue-bird:
```elm
String.fromList (String.toList str)
--> str

String.toList (String.fromList list)
--> list

-- all below are not in summary
String.toList << String.fromList
--> identity

String.fromList << String.toList
--> identity

String.toList << (String.fromList << f)
--> f

(f << String.toList) << String.fromList
--> f

String.fromList << (String.toList << f)
--> f

(f << String.fromList) << String.toList
--> f
```
and all variations with `<<`/`>>`

Not fully confident that the error messages are easy to understand/read.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -3418,10 +3418,21 @@ inversesCompositionCheck inverseFn checkInfo =
                 _ ->
                     Nothing
     in
-    firstThatConstructsJust
-        [ \() ->
-            case ( AstHelpers.getValueOrFunction checkInfo.earlier, AstHelpers.getComposition checkInfo.later ) of
-                ( Just earlierFnInfo, Just composition ) ->
+    case ( AstHelpers.getValueOrFunction checkInfo.earlier, AstHelpers.getValueOrFunction checkInfo.later ) of
+        ( Just earlierFnInfo, Just laterFnInfo ) ->
+            checkFnInfosForInverses
+                { earlierFnInfo = earlierFnInfo
+                , laterFnInfo = laterFnInfo
+                , details = [ "You can replace this composition by identity." ]
+                , fix =
+                    [ Fix.replaceRangeBy checkInfo.parentRange
+                        (qualifiedToString (qualify ( [ "Basics" ], "identity" ) checkInfo))
+                    ]
+                }
+
+        ( Just earlierFnInfo, Nothing ) ->
+            case AstHelpers.getComposition checkInfo.later of
+                Just composition ->
                     case AstHelpers.getValueOrFunction composition.earlier of
                         Just laterFnInfo ->
                             checkFnInfosForInverses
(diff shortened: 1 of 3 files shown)


src/Simplify.elm
```diff
@@ -3278,6 +3305,105 @@ consChecks checkInfo =
...
+                                        ++ removeAndBetweenFix { remove = Node.range composition.later, keep = Node.range composition.earlier }
+                                }
+
+                        Nothing ->
+                            Nothing
+
+                _ ->
+                    Nothing
+        , \() ->
+            case ( AstHelpers.getValueOrFunction checkInfo.earlier, AstHelpers.getValueOrFunction checkInfo.later ) of
```
jfmengels, line 3461: Should we extract these `getCompsoition` and `getValueOrFunction` calls to variables to avoid computing them multiple times?

tests/SimplifyTest.elm
```diff
@@ -6263,6 +6264,119 @@ a = []
...
+                            , under = "String.toList"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = (f)
+"""
+                        ]
+        , test "should replace String.toList << (f >> String.fromList) by (f)" <|
+            \() ->
+                """module A exposing (..)
+a = String.toList << (f >> String.fromList)
```
jfmengels, line 6332: This is really really nice :heart:

tests/SimplifyTest.elm
```diff
@@ -6466,6 +6482,22 @@ a = List.singleton >> String.fromList
...
+"""
+                    |> Review.Test.run (rule defaults)
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "String.toList and String.fromList cancel each other out"
+                            , details = [ "You can replace this call by the argument given to String.toList." ]
+                            , under = "String.fromList"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = (x |> f)
```
jfmengels, line 6503: Not sure why parens are introduced here :thinking:

jfmengels (approved):

jfmengels:
You have conflicts. You need to remove the delete the first four commits.

```bash
# Get the latest `main` branch
git checkout <your branch>
git rebase --interactive main
# then remove the first four commits (which are already on `main`)
# Save
git push --force <your branch>
```

tests/SimplifyTest.elm
lue-bird: fixed.

src/Simplify.elm
lue-bird: fixed.

lue-bird:
It's really strange because running `git rebase --interactive upstream/main` shows every commit except these 4. I'll try to close this PR and reopen first. If that doesn't work, I'll need some help.

jfmengels:
Hmm, yeah that's weird. I tend to avoid rebasing or doing things compared to remote branches. For instance, I prefer doing things operations relative to `main` rather than to `origin/main`, mostly to avoid doing it one way then the other and messing up. Therefore I prefer making that sure `main` points to the same thing as `origin/main`.

I'm curious to know the difference it makes in this case :man_shrugging: (but don't go investigating, unless you want to know yourself)

Issue #184, opened 2023-09-23
Add Set/Dict.fromList (toList x) --> x


lue-bird:
Similar to #183 
```elm
Set.fromList (Set.toList x) --> x
Dict.fromList (Dict.toList x) --> x
```

PR #186, opened 2023-09-23
Add Set.fromList on Set.toList simplification

lue-bird:
Part of #184 
```elm
Set.fromList (Set.toList set) --> set
```
including composition checks.


jfmengels (approved):
Nice :heart:

PR #190, opened 2023-09-23
Add Array.fromList + Array.toList simplifications

jfmengels:
Adds `Array.fromList >> Array.toList` and `Array.toList >> Array.fromList` simplifications

Shamefully stolen from the work on #184 :grin: 

Fixes some of the items in #174

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -717,6 +717,12 @@ Destructuring using case expressions
     Array.fromList []
     --> Array.empty
 
+    Array.fromList (Array.toList array)
+    --> array
+
+    Array.toList (Array.fromList list)
+    --> list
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -717,6 +717,12 @@ Destructuring using case expressions
     Array.fromList []
     --> Array.empty
 
+    Array.fromList (Array.toList array)
+    --> str
```
lue-bird, line 721: str should be array

lue-bird (approved):

src/Simplify.elm
jfmengels: I was so close!

PR #191, opened 2023-09-23
Add Array.length simplifications

jfmengels:
Adds simplifications for `Array.length` #174

- [x] `Array.length Array.empty` -> `0`
- [x] `Array.length (Array.fromList [ a, b, c ])` -> `3`
- [x] `Array.length (Array.repeat 1 x)` -> `1`
- [x] `Array.length (Array.initialize 1 f)` -> `1`
- [x]  `Array.length (Array.repeat n x)` to `max 0 n`
- [x]  `Array.length (Array.initialize n f)` to `max 0 n`

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -6227,19 +6227,39 @@ arrayLengthOnArrayRepeatOrInitializeChecks checkInfo =
     case maybeCall of
         Just ( fnName, call ) ->
             let
-                maxFn : String
-                maxFn =
-                    qualifiedToString (qualify ( [ "Basics" ], "max" ) defaultQualifyResources)
+                replacement : { details : List String, fix : List Fix }
+                replacement =
+                    case Evaluate.getInt checkInfo call.firstArg of
+                        Just length ->
+                            let
+                                positiveLength : String
+                                positiveLength =
+                                    String.fromInt (max 0 length)
+                            in
+                            { details = [ "You can replace this call by " ++ positiveLength ++ "." ]
+                            , fix = [ Fix.replaceRangeBy checkInfo.parentRange positiveLength ]
+                            }
+
+                        Nothing ->
+                            let
+                                maxFn : String
+                                maxFn =
+                                    qualifiedToString (qualify ( [ "Basics" ], "max" ) defaultQualifyResources)
+                            in
+                            { details =
+                                [ "You can replace this call by " ++ maxFn ++ " 0 with the given length. " ++ maxFn ++ " 0 makes sure that negative given lengths return 0." ]
+                            , fix =
+                                keepOnlyFix { parentRange = checkInfo.parentRange, keep = Node.range call.firstArg }
+                                    ++ [ Fix.insertAt checkInfo.parentRange.start (qualifiedToString (qualify ( [ "Basics" ], "max" ) checkInfo) ++ " 0 ") ]
+                            }
             in
             Just
                 (Rule.errorWithFix
                     { message = qualifiedToString (qualify checkInfo.fn checkInfo) ++ " on an array created by " ++ qualifiedToString (qualify ( [ "Array" ], fnName ) defaultQualifyResources) ++ " with a given length will result in that length"
-                    , details = [ "You can replace this call by " ++ maxFn ++ " 0 with the given length. " ++ maxFn ++ " 0 makes sure that negative given lengths return 0." ]
+                    , details = replacement.details
                     }
                     checkInfo.fnRange
-                    (keepOnlyFix { parentRange = checkInfo.parentRange, keep = Node.range call.firstArg }
-                        ++ [ Fix.insertAt checkInfo.parentRange.start (qualifiedToString (qualify ( [ "Basics" ], "max" ) checkInfo) ++ " 0 ") ]
-                    )
+                    replacement.fix
                 )
 
         Nothing ->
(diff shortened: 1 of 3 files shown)


tests/SimplifyTest.elm
```diff
@@ -15448,6 +15449,201 @@ a = Array.empty
...
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "The length of the array is 0"
+                            , details = [ "The length of the array can be determined by looking at the code." ]
+                            , under = "Array.length"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+import Array
+a = 0
+"""
```
jfmengels: This error gets removed in the next commits so that we only have 1 error reported instead of 2.

src/Simplify.elm
```diff
@@ -7708,22 +7782,31 @@ arrayDetermineSize resources expressionNode =
...
                 Nothing ->
                     Nothing
         ]
         ()
 
 
+ignoreCollectionSizeIfEmpty : Int -> Maybe CollectionSize
+ignoreCollectionSizeIfEmpty n =
+    if n <= 0 then
+        Nothing
```
lue-bird, line 7804: Why not return `Exactly 0`? We know the resulting size will be zero.

jfmengels: Because then we get 2 errors reported for the same issue (see https://github.com/jfmengels/elm-review-simplify/pull/191#discussion_r1335041738)
We get one error for `Array.repeat -2 x` and we get one for `Array.repeat (Array.repeat -2 x)`, which is not a great experience. Happy to be convinced otherwise though.

src/Simplify.elm
```diff
@@ -6163,6 +6182,61 @@ wrapperRepeatChecks wrapper checkInfo =
...
+                        maxFn =
+                            qualifiedToString (qualify ( [ "Basics" ], "max" ) defaultQualifyResources)
+                    in
+                    Just
+                        (Rule.errorWithFix
+                            { message = "The length of this array is the argument given to " ++ qualifiedToString (qualify ( [ "Array" ], fnName ) defaultQualifyResources)
+                            , details =
+                                [ "This is creating an array of a given size n, to then determine the size."
+                                , "You can replace this expression to " ++ maxFn ++ " 0 n. Calling " ++ maxFn ++ " is to handle the case where n might be negative."
+                                ]
```
lue-bird, line 6228: Maybe change the error info to something more like the other errors, like
```elm
{ message = ... checkInfo.fn ++ " on an array created by " ... fnName ++ " with a given length will result in that length"
, details = [ "You can replace this call by " ++ maxFn ++ " 0 with the given length. " ++ maxFn ++ " 0 makes sure that negative given lengths return 0." ]
}
```

src/Simplify.elm
```diff
@@ -6163,6 +6182,61 @@ wrapperRepeatChecks wrapper checkInfo =
...
+                        |> Maybe.map (Tuple.pair "repeat")
+                , \() ->
+                    AstHelpers.getSpecificFunctionCall ( [ "Array" ], "initialize" ) checkInfo.lookupTable checkInfo.firstArg
+                        |> Maybe.map (Tuple.pair "initialize")
+                ]
+                ()
+    in
+    case maybeCall of
+        Just ( fnName, call ) ->
+            case Evaluate.getInt checkInfo call.firstArg of
```
lue-bird, line 6211: This feels a tiny bit unnecessary because `collectionSizeChecks` will always be called earlier.

lue-bird (approved):

src/Simplify.elm
jfmengels: This is again to prevent having 2 errors. Not sure how to better to do that :thinking:

src/Simplify.elm
lue-bird: Aren't we doing that for multiple errors already?
Simple example being `Set.size (Set.fromList [])` where we both want to simplify `Set.fromList []` to `Set.empty` and `Set.size ...` to `0`.

jfmengels: Alright. I'll change it :+1:

src/Simplify.elm
jfmengels: Will remove this check

jfmengels: I ended up removing the check but keeping the computation to improve the message and the fix (otherwise we ended up with code like `max 0 -1`)

src/Simplify.elm
```diff
@@ -6163,6 +6182,72 @@ wrapperRepeatChecks wrapper checkInfo =
...
+        Just ( fnName, call ) ->
+            let
+                replacement : { details : List String, fix : List Fix }
+                replacement =
+                    case Evaluate.getInt checkInfo call.firstArg of
+                        Just length ->
+                            let
+                                positiveLength : String
+                                positiveLength =
+                                    String.fromInt (max 0 length)
```
lue-bird, line 6219: But why not do this in `arrayDetermineLength`?

jfmengels: Oversight, this helped simplify the code. Keep calling me out, it works great! :smile: (just to clarify, this isn't sarcasm)

lue-bird (approved):
<3 looks good

jfmengels:
Thank you for the feedback and corrections! :heart:

Issue #188, opened 2023-09-23
Additional simplifications enabled by #138


lue-bird:
https://github.com/jfmengels/elm-review-simplify/pull/138 included a list of checks that were enabled (made simpler) by that PR but haven't been added, yet.
- [ ] add `mapWrapChecks listCollection` to `listMapChecks`

PR #194, opened 2023-09-23
Add List.map on singleton simplifications

lue-bird:
The last unimplemented simplification from https://github.com/jfmengels/elm-review-simplify/issues/188
```elm
List.map f [ a ]
--> [ f a ]

-- below are not shown in summary for example
List.map f (if cond then [ x ] else [ y ])
--> [ f (if cond then x else y) ]

List.map f << List.singleton
--> List.singleton << f
```

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -507,6 +507,9 @@ Destructuring using case expressions
     List.map identity list
     --> list
 
+    List.map f [ a ]
+    --> [ f a ]
+
     List.filter (always True) list
     --> list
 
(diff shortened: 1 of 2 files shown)


jfmengels:
Hmm, I feel a bit uneasy with this one actually. `List.singleton >> List.map f` -> `f >> List.singleton` (and the function call variants) feels very nice, but I'm thinking that `List.map f [ a ]` -> `[ f a ]` could be a bit annoying.

For instance if we imagine that `f` is a large lambda and `a` a large expression (like an if/case expression), then moving all that code can make it harder to read than it was before :thinking: 

That said, this situation will be quite rare, so I'm *okay* to try it out. Hopefully people who don't like it will either report it back as an issue.

What do you think?

lue-bird:
This applies to all kinds of fixes, e.g.
```elm
Maybe.map f (Just a) --> Just (f a)
```
To me this very much feels like a simplification that should always be suggested.
Refactoring a possible lambda or argument into a let for example is a job for the user I'd say.
```elm
[ complex
    a
    b
]
    |> List.map (\element -> { element = element, range = range })
--> simplify
[ complex
    a
    b
    |> (\element -> { element = element, range = range })
]
--> user for example
let
    element =
        complex
            a
            b
in
[ { element = element, range = range } ]
```

jfmengels:
Alright. Yeah, it doesn't look that bad, especially since you keep the pipelines. Alright :+1:

PR #199, opened 2023-09-24
Add Array.append simplifications

jfmengels:
Adds Array.append simplifications #174 

- [x] `Array.append Array.empty array` -> `array`
- [x] `Array.append array Array.empty` -> `array`
- [x] `Array.append (Array.fromList [ a, b, c ]) (Array.fromList [ d, e, f ])` -> `Array.fromList [ a, b, c, d, e, f ]`

I extracted `listAppendChecks` into a generic `collectionAppendChecks`. I noticed it got very close to `collectionUnionChecks` and changed the error messages to match `collectionUnionChecks` because they were nicer.

Now `collectionAppendChecks` and `collectionUnionChecks` are pretty much the same check, except that the former tries to combine `append [ a ] [ b ]` into `[ a, b ]`.

I think we can make that work for dict/set as well in a next step, completely fusing the 2 collection checks.
The only difference is that sets and dict need to have the elements added to the end rather than at the beginning, so `union (fromList [ a ]) (fromList [ b ])` into `fromList [ b, a ]`. I'll do that in a follow-up PR.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -765,6 +765,12 @@ Destructuring using case expressions
     Array.length (Array.initialize n f)
     --> max 0 n
 
+    Array.append Array.empty array
+    --> array
+
+    Array.append (Array.fromList [ a, b ]) (Array.fromList [ c, d ])
+    --> (Array.fromList [ a, b, c, d ])
+
 
 ### Sets
 
@@ -2503,6 +2509,7 @@ functionCallChecks =
         , ( ( [ "Array" ], "length" ), arrayLengthChecks )
         , ( ( [ "Array" ], "repeat" ), arrayRepeatChecks )
         , ( ( [ "Array" ], "initialize" ), arrayInitializeChecks )
+        , ( ( [ "Array" ], "append" ), collectionAppendChecks arrayCollection )
         , ( ( [ "Set" ], "map" ), emptiableMapChecks setCollection )
         , ( ( [ "Set" ], "filter" ), emptiableFilterChecks setCollection )
         , ( ( [ "Set" ], "remove" ), collectionRemoveChecks setCollection )
@@ -7804,7 +7811,7 @@ stringDetermineLength expression =
             Nothing
 
 
-arrayCollection : CollectionProperties {}
+arrayCollection : CollectionProperties (FromListProperties {})
 arrayCollection =
     { moduleName = [ "Array" ]
     , represents = "array"
@@ -7819,6 +7826,10 @@ arrayCollection =
         }
     , nameForSize = "length"
     , determineSize = arrayDetermineSize
+    , fromListLiteralRange =
+        \lookupTable expr ->
+            AstHelpers.getSpecificFunctionCall ( [ "Array" ], "fromList" ) lookupTable expr
+                |> Maybe.andThen (\{ firstArg } -> AstHelpers.getListLiteralRange firstArg)
     }
 
 
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -5192,7 +5192,7 @@ collectionAppendChecks collection checkInfo =
                                     Just literalListRangeFirst ->
                                         Just
                                             (Rule.errorWithFix
-                                                { message = "Appending literal " ++ collection.represents ++ "s could be simplified to be a single " ++ collection.represents
+                                                { message = qualifiedToString (qualify checkInfo.fn defaultQualifyResources) ++ " on literal " ++ collection.represents ++ "s can be turned into a single literal " ++ collection.represents
                                                 , details = [ "Try moving all the elements into a single " ++ collection.represents ++ "." ]
                                                 }
                                                 checkInfo.fnRange
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -2459,7 +2459,7 @@ functionCallChecks =
         , ( ( [ "Result" ], "andThen" ), resultAndThenChecks )
         , ( ( [ "Result" ], "withDefault" ), withDefaultChecks resultWithOkAsWrap )
         , ( ( [ "Result" ], "toMaybe" ), unwrapToMaybeChecks resultWithOkAsWrap )
-        , ( ( [ "List" ], "append" ), listAppendChecks )
+        , ( ( [ "List" ], "append" ), collectionAppendChecks listCollection )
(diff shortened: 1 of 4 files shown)


src/Simplify.elm
```diff
@@ -5156,62 +5163,59 @@ listIntersperseCompositionChecks checkInfo =
...
-                                    { arg = listArgToTheLeft
-                                    , argRepresents = "first list argument"
-                                    }
-                                    checkInfo
-                                )
+                Just secondArg_ ->
+                    if collection.empty.is checkInfo.lookupTable secondArg_ then
+                        Just
+                            (Rule.errorWithFix
+                                { message = "Unnecessary append with " ++ descriptionForIndefinite collection.empty.description
```
lue-bird, line 5188: I'd prefer `qualifiedToString checkInfo.fn` instead of just "append". Should be more generic (e.g. if append is called differently e.g. concat or union).

src/Simplify.elm
```diff
@@ -5156,62 +5163,59 @@ listIntersperseCompositionChecks checkInfo =
...
-                                            ]
-                                        )
+                    else
+                        case collection.fromListLiteralRange checkInfo.lookupTable secondArg_ of
+                            Just literalListRangeSecond ->
+                                case collection.fromListLiteralRange checkInfo.lookupTable checkInfo.firstArg of
+                                    Just literalListRangeFirst ->
+                                        Just
+                                            (Rule.errorWithFix
+                                                { message = "Appending literal " ++ collection.represents ++ "s could be simplified to be a single " ++ collection.represents
```
lue-bird, line 5202: As above, I'd prefer using `... checkInfo.fn ++ " on ..."` instead of `"Appending ..."` to be more precise and situational.
Also not sure I'd call `Array.fromList [ 1, 2, 3 ]` a "literal".

tests/SimplifyTest.elm
```diff
@@ -16044,6 +16047,240 @@ a = max 0 n
...
+arrayAppendTests : Test
+arrayAppendTests =
+    describe "Array.append"
+        [ test "should not report Array.append with an array variable" <|
+            \() ->
+                """module A exposing (..)
+import Array
+a = Array.append
+b = Array.append [ 1 ]
+c = Array.append [ 1 ] array
```
lue-bird, line 16059: append on `[ 1 ]` is a compiler error. Use `Array.fromLost [ 1 ]`?

lue-bird:
This has been on my wishlist for some time. Very cool!

src/Simplify.elm
```diff
@@ -7455,6 +7459,12 @@ type alias WrapperProperties otherProperties =
         }
 
 
+type alias FromListProperties otherProperties =
+    { otherProperties
```
lue-bird, line 7463: All other `-Properties` types start by requiring the default `TypeProperties`. Maybe add this here, too, for consistency?
```elm
type alias FromListProperties otherProperties =
    TypeProperties
        { otherProperties | ... }
```

src/Simplify.elm
jfmengels: Yeah, was going to do this in the next PR.
I agree the term "literal" is not perfect, but I think it's what I think comes closest. Let's use this until we come up with something better.

src/Simplify.elm
jfmengels: Done :+1:

src/Simplify.elm
lue-bird: I would not just use `Tuple.second` (which btw should probably be `AstHelpers.qualifiedName`) as just the name here. Why not use the qualified name?

jfmengels: It was to keep the same error message as for the union checks and the list append checks. But yeah, let's change it :+1:

src/Simplify.elm
lue-bird: I would not just use `Tuple.second` as only the unqualified name here (which btw should probably be `AstHelpers.qualifiedName`). Why not use the qualified name?

src/Simplify.elm
```diff
@@ -5156,62 +5163,59 @@ listIntersperseCompositionChecks checkInfo =
...
-                                            ]
-                                        )
+                    else
+                        case collection.fromListLiteralRange checkInfo.lookupTable secondArg_ of
+                            Just literalListRangeSecond ->
+                                case collection.fromListLiteralRange checkInfo.lookupTable checkInfo.firstArg of
+                                    Just literalListRangeFirst ->
+                                        Just
+                                            (Rule.errorWithFix
+                                                { message = qualifiedToString (qualify checkInfo.fn defaultQualifyResources) ++ " on literal " ++ collection.represents ++ "s can be turned into a single " ++ collection.represents
```
lue-bird, line 5202: "... on literal arrays can be turned into a single array" sounds a bit weird. Why not reuse the term literal? "can be turned into a single literal array"

jfmengels: Sounds good :+1:

lue-bird:
I suspect there could be issues with cases like
```elm
Array.append ([ 1 ] |> Array.fromList) ([ 2 ] |> Array.fromList)
```
Would be cool to add tests for these cases

jfmengels:
I added more tests, it looks like things are working as expected :+1:

src/Simplify.elm
```diff
@@ -7814,6 +7826,10 @@ arrayCollection =
         }
     , nameForSize = "length"
     , determineSize = arrayDetermineSize
+    , fromListLiteralRange =
+        \lookupTable expr ->
+            AstHelpers.getSpecificFunctionCall ( [ "Array" ], "fromList" ) lookupTable expr
+                |> Maybe.andThen (\{ firstArg } -> AstHelpers.getListLiteralRange firstArg)
```
lue-bird, line 7832: All looks good!

(I _personally_ dislike record destructuring because of potential future name clashes etc, so I always prefer
```elm
|> Maybe.andThen (\call -> AstHelpers.getListLiteralRange call.firstArg)
```
instead of
```elm
|> Maybe.andThen (\{ firstArg } -> AstHelpers.getListLiteralRange firstArg)
```
but no need to change this now. Should be fine)

lue-bird (approved):

src/Simplify.elm
jfmengels: Sure. Will change it in a following PR.

PR #201, opened 2023-09-24
Combine literals when using Dict/Set.union

jfmengels:
Adds simplifications made possible in #199 for `Array.append` for `Dict.union` and `Set.union`.

```elm
Set.union (Set.fromList [ a, b ]) (Set.fromList [ c, d ])
--> Set.fromList [ a, b, c, d ]

-- Note: Order is reversed
Dict.union (Dict.fromList [ a, b ]) (Dict.fromList [ c, d ])
--> Dict.fromList [ a, b, c, d ]
```

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -2471,7 +2471,7 @@ functionCallChecks =
         , ( ( [ "Result" ], "andThen" ), resultAndThenChecks )
         , ( ( [ "Result" ], "withDefault" ), withDefaultChecks resultWithOkAsWrap )
         , ( ( [ "Result" ], "toMaybe" ), unwrapToMaybeChecks resultWithOkAsWrap )
-        , ( ( [ "List" ], "append" ), collectionAppendChecks listCollection )
+        , ( ( [ "List" ], "append" ), collectionUnionChecks listCollection )
         , ( ( [ "List" ], "head" ), listHeadChecks )
         , ( ( [ "List" ], "tail" ), listTailChecks )
         , ( ( [ "List" ], "member" ), listMemberChecks )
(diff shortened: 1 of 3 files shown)


src/Simplify.elm
```diff
@@ -813,6 +813,9 @@ Destructuring using case expressions
     Set.union set Set.empty
     --> set
 
+    Set.union (Set.fromList [ a, b ]) (Set.fromList [ c, d ])
+    --> Set.fromList [ a, b, c, d]
```
lue-bird, line 817: Do `Set.union` and `Dict.union` really behave differently?
Looking at their implementation, it doesn't seem like it: https://github.com/elm/core/blob/1.0.5/src/Set.elm#L100

jfmengels: No, it's more that `Dict.union` needs to have things inserted in the opposite order than `Array.append`/`List.append`, otherwise you ge tbehavior changes if you have
```elm
Dict.union (Dict.fromList [ ( "a", 1 ) ]) (Dict.fromList [ ( "a", 2 ) ])
-- results in Dict.fromList [ ( "a", 1 ) ]
-- so we should simplify it to
--> Dict.fromList [ ( "a", 2 ), ( "a", 1 ) ]

-- If we simplified it to
Dict.fromList [ ( "a", 1 ), ( "a", 2 ) ]
-- then the value of that would be 
-- Dict.fromList [ ( "a", 2 ) ]
```
So we need to reverse it.

`Set` behaves just like `Dict`, except that it is is a special case of `Dict` where the order really doesn't matter. So I chose to keep the same behavior as `Array.append`/`List.append` which I think feels a lot more natural.

lue-bird (approved):

PR #202, opened 2023-09-24
Add String.append simplifications

jfmengels:
Same as #199 but for Strings.

I also made
```elm
String.append "" str
--> str

String.append (String.fromList [ a, b ]) (String.fromList [ c, d ])
--> String.fromList [ a, b, c, d ]
```

The same simplification now also works with `++`:

```elm
String.fromList [ a, b ] ++ String.fromList [ c, d ]
--> String.fromList [ a, b, c, d ]
```
and `++` with regular lists now (partially) uses the same check.

### DIFF
--- a/tests/SimplifyTest.elm
+++ b/tests/SimplifyTest.elm
@@ -5737,6 +5737,7 @@ stringSimplificationTests =
         , stringReplaceTests
         , stringWordsTests
         , stringLinesTests
+        , stringAppendTests
         , stringReverseTests
         , stringSliceTests
         , stringRightTests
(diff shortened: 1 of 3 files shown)


tests/SimplifyTest.elm
```diff
@@ -5717,6 +5717,22 @@ a = left ++ ([ b ] ++ c)
...
                     |> Review.Test.expectNoErrors
+        , test "should replace String.fromList [ b, c ] ++ String.fromList [ d, e ] by String.fromList [ b, c , d, e ]" <|
+            \() ->
+                """module A exposing (..)
+a = String.fromList [ b, c ] ++ String.fromList [ d, e ]
+"""
+                    |> Review.Test.run ruleWithDefaults
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "++ on literal strings can be turned into a single literal string"
```
lue-bird, line 5728: Here we kind of see why literal isn't a perfect name for this. A string literal just isn't the same as "String.fromList on a list literal".
Maybe we can add something like `fromListLiteralDescription : String` to the `FromListProperties` and use that?

jfmengels: Yeah, that sounds good 👍

jfmengels: I decided to go with "X.fromList call" to clear out the confusion. Would "X.fromList construction" (or something else) be better? 🤔

lue-bird: That sounds so, so much better. Good job!

PR #203, opened 2023-09-25
Add simplifications for Array.get

jfmengels:
Adds the following simplifications:

- [x] `Array.get n Array.empty` -> `Nothing`
- [x] `Array.get -3 array` -> `Nothing`
- [x] `Array.get 2 (Array.fromList [ 1, 2, 3 ])` -> `Just 3`
- [x] `Array.get 200 (Array.fromList [ 1, 2, 3 ])` -> `Nothing`

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -780,6 +780,9 @@ Destructuring using case expressions
     Array.get n Array.empty
     --> Nothing
 
+    Array.get 1 (Array.fromList [ a, b, c ])
+    --> Just b
+
     Array.get -1 array
     --> Nothing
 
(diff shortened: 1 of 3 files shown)


src/Simplify.elm
```diff
@@ -6266,25 +6270,84 @@ arrayGetChecks checkInfo =
...
+                                    just =
+                                        qualifiedToString (qualify ( [ "Maybe" ], "Just" ) checkInfo)
+                                in
+                                Just
+                                    (Rule.errorWithFix
+                                        { message = "The element returned by " ++ qualifiedToString checkInfo.fn ++ " is known"
+                                        , details = [ "You can replace this call by Just the targeted element." ]
+                                        }
+                                        checkInfo.fnRange
+                                        (keepOnlyFix { parentRange = checkInfo.parentRange, keep = Node.range element }
```
lue-bird, line 6309: why not `replaceBySubExpression` and remove the paren insertion?

src/Simplify.elm
```diff
@@ -6339,7 +6342,14 @@ indexAccessChecks collection checkInfo n =
                                     )
 
                             Nothing ->
-                                Nothing
+                                Just
+                                    (Rule.errorWithFix
+                                        { message = "The index is out of the bounds of the given " ++ collection.represents
```
lue-bird, line 6347: Why not add the specific function as we do in all other errors? checkInfo.fn ++ " with an index out of the bounds of the given " +++ represents

src/Simplify.elm
```diff
@@ -6246,6 +6260,104 @@ arrayLengthOnArrayRepeatOrInitializeChecks checkInfo =
             Nothing
 
 
+arrayGetChecks : CheckInfo -> Maybe (Error {})
```
lue-bird, line 6263: Maybe make this generic for all `EmptiableProperties (IndexableProperties _)`?

src/Simplify.elm
jfmengels: Will do, that sounds better.

src/Simplify.elm
jfmengels: Yeah, that was much nicer. At the moment, I find it hard to know which checks, fixes and other helpers to reach for. I try to look at how similar functions do things and I try to find things by name, but I often still miss things. I originally copied this from somewhere (and then added the parens insertion).

lue-bird: Would be nice to have a local `Fix.Extra` with all our nice fix helpers maybe.

PR #204, opened 2023-09-25
Limit the number of args to what each check expects

jfmengels:
This will help avoid fixes from removing too many arguments which often happens for instance when replacing the entire expression.

I noticed this could be a problem for instance when dealing with `List.foldl f initial [] unexpectedArgument` which we would fix as `initial` instead as of `initial unexpectedArgument` (because the result of the fold function could be a new function).

I'm limiting the number of args each check expects to what the function to check expects.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1910,12 +1910,40 @@ expressionVisitorHelp (Node expressionRange expression) config context =
         toCheckInfo :
             { fnRange : Range
             , fn : ( ModuleName, String )
+            , argCount : Int
             , firstArg : Node Expression
             , argsAfterFirst : List (Node Expression)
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -1910,12 +1910,40 @@ expressionVisitorHelp (Node expressionRange expression) config context =
         toCheckInfo :
             { fnRange : Range
             , fn : ( ModuleName, String )
+            , nbOfArgs : Int
```
lue-bird, line 1913: Adding this makes _so_ much sense!
Maybe call it `argCount`?

jfmengels: Sure

Issue #154, opened 2023-09-16
Reverse composition


jfmengels:
Seems like we don't simplify the following things
- `List.reverse >> List.reverse` --> `identity`
- `String.reverse >> String.reverse` --> `identity`

PR #205, opened 2023-09-25
Add String.reverse, List.reverse composition simplifications

lue-bird:
Implements #154. Only 2 code lines added, so having both in one PR seemed easiest.


jfmengels (approved):
Easy addition :grin:

PR #207, opened 2023-09-26
Add Array.set simplifications

jfmengels:
Adds the follow `Array` simplifications from #174:

- [x] `Array.set n x Array.empty` -> `Array.empty`
- [x] `Array.set 1 x (Array.fromList [ 1, 2, 3 ])` -> `Array.fromList [ 1, x, 3 ]`
- [x] `Array.set 100 x (Array.fromList [ 1, 2, 3 ])` -> `Array.fromList [ 1, 2, 3 ]`
- [x] `Array.set -1 x array` -> `array`

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -6426,8 +6426,8 @@ setOnKnownElementChecks collection checkInfo n replacementArgRange =
                         Just element ->
                             Just
                                 (Rule.errorWithFix
-                                    { message = qualifiedToString checkInfo.fn ++ " will replace a known element in a literal " ++ collection.represents
-                                    , details = [ "You can move the replacement argument directly into the " ++ collection.represents ++ "." ]
+                                    { message = qualifiedToString checkInfo.fn ++ " will replace a known element in the " ++ collection.fromListLiteralDescription
+                                    , details = [ "You can move the replacement argument directly into the " ++ collection.fromListLiteralDescription ++ "." ]
                                     }
                                     checkInfo.fnRange
                                     (keepOnlyFix { parentRange = checkInfo.parentRange, keep = Node.range collectionArg }
--- a/tests/SimplifyTest.elm
+++ b/tests/SimplifyTest.elm
@@ -16990,8 +16990,8 @@ a = Array.set 1 x (Array.fromList [ b, c, d ])
                     |> Review.Test.run ruleWithDefaults
                     |> Review.Test.expectErrors
                         [ Review.Test.error
-                            { message = "Array.set will replace a known element in a literal array"
-                            , details = [ "You can move the replacement argument directly into the array." ]
+                            { message = "Array.set will replace a known element in the Array.fromList call"
+                            , details = [ "You can move the replacement argument directly into the Array.fromList call." ]
                             , under = "Array.set"
                             }
                             |> Review.Test.whenFixed """module A exposing (..)


src/Simplify.elm
```diff
@@ -6370,6 +6383,89 @@ indexAccessChecks collection checkInfo n =
...
+setOnKnownElementChecks collection checkInfo n replacementArgRange =
+    case thirdArg checkInfo of
+        Just collectionArg ->
+            case collection.literalElements checkInfo.lookupTable collectionArg of
+                Just literalElements ->
+                    case List.drop n literalElements |> List.head of
+                        Just element ->
+                            Just
+                                (Rule.errorWithFix
+                                    { message = "The element returned by " ++ qualifiedToString checkInfo.fn ++ " is known"
```
lue-bird, line 6443: This message seems to be copied from get but it is both incorrect and very undescriptive I'd say.

lue-bird (changes requested):

src/Simplify.elm
```diff
@@ -6370,6 +6383,89 @@ indexAccessChecks collection checkInfo n =
...
+
+                        Nothing ->
+                            Just
+                                (Rule.errorWithFix
+                                    { message = qualifiedToString checkInfo.fn ++ " with negative index will always return the same given " ++ collection.represents
+                                    , details =
+                                        [ "You can replace this call by \\_ -> identity." ]
+                                    }
+                                    checkInfo.fnRange
+                                    [ Fix.replaceRangeBy checkInfo.parentRange "\\_ -> identity" ]
```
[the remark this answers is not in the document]
jfmengels: All good points :+1: 
For the `if`, it used to be in the correct location, and then I changed it. Seems like I'm missing a test case, so that's good to discover.
(And doh, didn't think of using `always`  *face-palm*)

src/Simplify.elm
```diff
@@ -10359,26 +10352,52 @@ Use `returnsArgError` with the given last arg as `arg` when the last arg is alre
...
+                            , fix =
+                                [ Fix.replaceRangeBy resources.parentRange
+                                    (qualifiedToString (qualify ( [ "Basics" ], "identity" ) resources))
+                                ]
+                            }
+
+                        2 ->
+                            { description = "always"
+                            , fix =
+                                [ Fix.replaceRangeBy resources.parentRange
```
lue-bird, line 10383: should be `always identity`

src/Simplify.elm
```diff
@@ -10359,26 +10352,52 @@ Use `returnsArgError` with the given last arg as `arg` when the last arg is alre
...
+                                    (qualifiedToString (qualify ( [ "Basics" ], "always" ) resources))
+                                ]
+                            }
+
+                        _ ->
+                            -- Use-case is absent for now
+                            { description = "the " ++ config.lastArgRepresents ++ " argument"
+                            , fix = []
```
lue-bird, line 10391: should be `"(\\ " ++ repeat remainingArgCount "_ " ++ " -> identity)`

src/Simplify.elm
```diff
@@ -10359,26 +10352,52 @@ Use `returnsArgError` with the given last arg as `arg` when the last arg is alre
 alwaysReturnsLastArgError :
     String
     ->
-        { lastArgRepresents : String
-        , lastArg : Maybe (Node Expression)
+        { remainingArgCount : Int
+        , remainingArgs : List (Node Expression)
```
lue-bird, line 10356: why not use `argCount` and `argsAfterFirst` from checkInfo to determine the level of currying?

src/Simplify.elm
lue-bird:
```elm
> Array.set -1
<function> : a -> Array.Array a -> Array.Array a
> always identity
<function> : b -> a -> a
```
but
```elm
> always
<function> : a -> b -> a
```
does not unify (the element type and array type aren't necessarily the same)

src/Simplify.elm
jfmengels: Because it's not always that number of arguments.
Example (https://github.com/jfmengels/elm-review-simplify/pull/207/files#diff-6239bccafe2b75c7e560542732acf25d142d08608e1e0f0ffe9b53c09cfb1e0eR4837):
```elm
String.replace n n
```
Here we're only expecting 1 more argument because we already have one.

src/Simplify.elm
jfmengels: Should be "(\\ " ++ repeat (remainingArgCount - 1) "_ " ++ " -> identity)` in that case I think.

- No missing argument: take the last argument
- 1 missing argument: `identity`
- 2 missing arguments: `always`
- 3 missing arguments: `\_ _ -> identity`?

Feels a bit inconsistent, no?

src/Simplify.elm
lue-bird: `argsAfterFirst` tells you exactly how many are actually present, right?
So together with `argCount` we already have all the info we need to determine both the last arg and how many args before which aren't present need to be ignored

jfmengels: Hmm... yeah you're right, they're equivalent in practice :+1:

src/Simplify.elm
jfmengels: I'll use `always identity` for now, we can always revisit this later.

src/Simplify.elm
```diff
@@ -6370,6 +6379,79 @@ indexAccessChecks collection checkInfo n =
...
+    case thirdArg checkInfo of
+        Just collectionArg ->
+            case collection.literalElements checkInfo.lookupTable collectionArg of
+                Just literalElements ->
+                    case List.drop n literalElements |> List.head of
+                        Just element ->
+                            Just
+                                (Rule.errorWithFix
+                                    { message = qualifiedToString checkInfo.fn ++ " will replace a known element in a literal " ++ collection.represents
+                                    , details = [ "You can move the replacement argument directly into the " ++ collection.represents ++ "." ]
```
lue-bird, line 6430: Instead of using `collection.represents` in both places, why not use `collection.fromListLiteralDescription`?

lue-bird (approved):

jfmengels:
Thank you for your help!

Issue #209, opened 2023-09-27
Detect parenthesized composition


lue-bird:
This is more a note to self for the future:
Currently, only `inversesCompositionCheck`, `toggleCompositionChecks` support reporting the first 2 in
```elm
(f >> earlier) >> later
earlier >> (later >> g)
(f >> earlier) >> (later >> g)
-- even with infinitely nested parens
(f >> (g >> earlier)) >> later
earlier >> ((later >> f) >> g)
(f >> (g >> earlier)) >> ((later >> h) >> i)
```
with any combination of `>>` and `<<`.

In practice, this means adding `earlier.removeRange`, `later.removeRange` to `CompositionIntoChecks` and re-checking that the fixes work using these.

Advantages
- reporting more
- `inversesCompositionCheck` and `toggleCompositionChecks` can be part of `compositionIntoChecks`, something that has been bugging me

jfmengels:
I think that for `(f >> earlier) >> later`, we could simply have a fix that reports the parentheses. That would be a simplification in itself too.

If we do that, I think it could still make sense to improve our support/understanding of `later << (f >> earlier)`, although a case could be made that the order should be made that that should be simplified. Though I don't think this rule should/could tell which composition operator should be preferred.

Issue #162, opened 2023-09-17
Record access simplifications


jfmengels:
We already have
```elm
{ field = 1 }.field
--> 1
```
but we don't have
```elm
.field { field = 1 }
--> 1
```

I think we can take the same simplifications done in #161 for tuples and apply them for records where we currently don't

(✅ indicates this is now supported)

```elm
-- already shown above
.field { field = a } ✅
--> a

.field { record | field = a } ✅
--> a

.field { record | otherField = a } ✅
--> record.field

.field << (\a -> { field = b, other = other }) ✅
--> (\a -> b)

.field << (\a -> { record | field = b, other = other }) ✅
--> (\a -> b)

.field << (\a -> { record | otherField = a }) ✅
--> .field << (\a -> record)

(.field >> .bar) << (\a -> { record | otherField = a }) ✅
--> (.field >> bar) << (\a -> record)

-- Not this one, unless we want to inline `.field`, in which case we could do
.field << (\a -> if condition then { field = b, other = other } else { record | otherField = a })
--> (\a -> if condition then b else record.field)
```

And then if we the names and positions of record fields of type aliases, we can do the exact same simplifications done in #161 for tuples when using the automatic record constructor.

```elm
type alias Record = { first : Int, second : Int }

(Record a b).first ✅
--> a

.first (Record a b) ✅
--> a

.second (Record a b) ✅
--> b

.second << Record a
--> identity

.first << Record a
--> always a
```

lue-bird:
These 2 don't make sense to me
```elm
.field << (\a -> { field = a, other = other })
--> a

.field << (\a -> { record | field = a, other = other })
--> a
```
I think both should return `(\a -> a)` (or `identity`), right?

jfmengels:
Right! 👍
Probably the lambda version, because that applies to more situations. (Will edit the proposals when back on my computer)

jfmengels:
@lue-bird Shall I move the proposals for `type alias Record = { first : Int, second : Int }` into a separate issue?

PR #210, opened 2023-09-27
Add compose into record access of known field checks

lue-bird:
Completes #162 except the parenthesized composition which will be done together with #209 
```elm
.field << (\a -> { field = b, other = other })
--> (\a -> b)

.field << (\a -> { record | field = b, other = other })
--> (\a -> b)

.field << (\a -> { record | otherField = b })
--> (\a -> record.field)
```
Note that your original proposal was
```elm
.field << (\a -> { record | otherField = b })
--> .field << (\a -> record)
```
If you prefer that, I'll change it.


jfmengels:
I think your simplication version is very nice and will likely lead to better simplifications :+1:

Issue #155, opened 2023-09-16
Redundant sorting


jfmengels:
`List.sort >> List.sort` is redundant and can be simplified to `List.sort`.

---

For `List.sortWith f >> List.sortWith g`, logically we could remove `List.sortWith f`, as whatever ordering `f` does gets overriden by `g`, but that's not always true, so we should not do this (unless we check and notice that `f == g`).

Counter-example where the double sorting has an effect (using `sortBy` which is similar):
```elm
data = [ ( 2, 2), ( 1, 100 ), ( 1, 1 ) ]

data |> List.sortBy Tuple.second |> List.sortBy Tuple.first
--> [(1,1),(1,100),(2,2)]

data |> List.sortBy Tuple.first
--> [(1,100),(1,1),(2,2)]
```

PR #212, opened 2023-09-30
Add sort on equivalent sort call simplifications

lue-bird:
Implements #155 and add the same checks for `sortBy`, `sortWith` with equal arguments

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -2794,6 +2794,8 @@ compositionIntoChecks =
         , ( ( [ "Result" ], "toMaybe" ), resultToMaybeCompositionChecks )
         , ( ( [ "List" ], "reverse" ), listReverseCompositionChecks )
         , ( ( [ "List" ], "sort" ), listSortCompositionChecks )
+        , ( ( [ "List" ], "sortBy" ), listSortByCompositionChecks )
+        , ( ( [ "List" ], "sortWith" ), listSortWithCompositionChecks )
         , ( ( [ "List" ], "map" ), listMapCompositionChecks )
         , ( ( [ "List" ], "filterMap" ), listFilterMapCompositionChecks )
         , ( ( [ "List" ], "intersperse" ), listIntersperseCompositionChecks )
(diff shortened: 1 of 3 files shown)


src/Simplify.elm
```diff
@@ -6445,10 +6453,110 @@ listSortChecks checkInfo =
...
         ]
         ()
 
 
+listSortCompositionChecks : CompositionIntoCheckInfo -> Maybe ErrorInfoAndFix
+listSortCompositionChecks checkInfo =
+    operationDoesNotChangeResultOfOperationCompositionCheck { argCount = 1 } checkInfo
+
+
+{-| Condense applying the the same function with equal arguments (except the last one) twice in sequence into one.
```
jfmengels, line 6466:
```suggestion
{-| Condense applying the same function with equal arguments (except the last one) twice in sequence into one.
```

src/Simplify.elm
```diff
@@ -6445,10 +6453,110 @@ listSortChecks checkInfo =
...
+This applies to functions that are equivalent to identity when operating on the result another such function.
+
+Examples of such functions:
+
+  - one argument: `Simplify.expectNaN`, `Review.Rule.providesFixesForModuleRule`, `List.sort`, `List.Extra.unique`, [`AVL.Set.clear`](https://package.elm-lang.org/packages/owanturist/elm-avl-dict/2.1.0/AVL-Set#clear)
+  - two arguments: `List.filter f`, `List.take/drop n`, `List.Extra.filterNot f`, `List.Extra.takeWhile/dropWhile(Right) f`, `List.sortBy f`, `List.sortWith f`, `List.Extra.uniqueBy f`
```
jfmengels, line 6472: `List.drop n >> List.drop n` results in `List.drop (n + n)`, so it's not a good example. (But it could be a new simplification idea), I'll add it to #2.

src/Simplify.elm
lue-bird: thanks, fixed.

src/Simplify.elm
lue-bird: very true. My mind was in `drop/takeWhile` land. Fixed.

jfmengels (approved):

jfmengels:
Nice :blush:

PR #213, opened 2023-09-30
Add String.concat on List.repeat call simplifications

lue-bird:
Adds the last `String` simplification from #2 
```elm
String.concat (List.repeat n str)
--> String.repeat n str
```
including composition.

Bonus: Corrected test names.

I'm thinking we could add the same for `String.concat (List.intersperse s strings) --> String.join s strings`


jfmengels (approved):
Nice! I imagine the intent is to have these new utilities make it to our other existing simplifications?

jfmengels:
> I'm thinking we could add the same for String.concat (List.intersperse s strings) --> String.join s strings

That sounds good :+1

lue-bird:
> Nice! I imagine the intent is to have these new utilities make it to our other existing simplifications?

Exactly right.

PR #217, opened 2023-10-01
Add Set.foldl/r simplifications

lue-bird:
Including parts from #2
```elm
-- The following simplifications for Set.foldl also work for Set.foldr
Set.foldl f initial Set.empty
--> initial
Set.foldl (\_ soFar -> soFar) initial set
--> initial
```

The added helpers are generic enough to be used for `Dict/Array/String/List.foldlr` as well

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 - `List.sortBy f (List.sortBy f list)` to `List.sortBy f list`
 - `String.concat (List.repeat n str)` to `String.repeat n str`
 - `String.concat (List.intersperse str strings)` to `String.join str strings`
+- `Set.foldl/r f initial Set.empty` to `initial`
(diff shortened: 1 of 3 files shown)


jfmengels (approved):
LGTM. I think the tests are failing because of `elm-format`. Do you also have problems for running it in the tests file? I do at least, and I'm guessing it's because the file is getting too big.

lue-bird:
No, it runs fine. I just forgot it :)

jfmengels:
Wait, you don't run it on save? :scream:

lue-bird:
I do run it on save on all other projects but in simplify I use the elm-land one which sometimes gets tripped up when formatting.

jfmengels:
Aha, alright. Worth a bug report there probably :thinking:

PR #230, opened 2023-10-02
Add Maybe.mapN simplifications

lue-bird:
`Maybe.mapN` simplifications from #2 
```elm
-- the following simplifications for map2 work for all Maybe.mapN
Maybe.map2 f firstMaybe Nothing --> Nothing
Maybe.map2 f (Just a) (Just b) --> Just (f a b)
```


jfmengels (approved):
:heart:

PR #232, opened 2023-10-02
Add Array.toIndexedMap simplifications

lue-bird:
Ad the `Array.toIndexedMap` simplifications listed in #174 
```elm
Array.toIndexedList Array.empty
--> []

List.map Tuple.second (Array.toIndexedList array)
--> Array.toList array
```
including composition

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -5758,12 +5758,12 @@ arrayToIndexedListMapCompositionCheck : CompositionIntoCheckInfo -> Maybe ErrorI
 arrayToIndexedListMapCompositionCheck checkInfo =
     case ( checkInfo.earlier.fn, checkInfo.later.args ) of
         ( ( [ "Array" ], "toIndexedList" ), elementMappingArg :: [] ) ->
-            let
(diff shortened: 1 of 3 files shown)


src/Simplify.elm
```diff
@@ -853,6 +853,12 @@ Destructuring using case expressions
     Array.foldl (\_ soFar -> soFar) initial array
     --> initial
 
+    Array.toIndexedList Array.empty
+    --> []
+
+    List.map Tuple.second (Array.toIndexedList array)
+    --> Array.toList array
```
jfmengels, line 860: Do you think we should also have this?
```elm
List.map Tuple.first (Array.toIndexedList array)
--> List.range 0 (Array.length array - 1)
```

lue-bird: Would you want to add it? It should be more performant at least.

jfmengels: I think it makes sense yeah. It's a bit simpler, shorter, more performant and it enables more optimizations with the `Array.length` checks we have.

PR #233, opened 2023-10-02
Simplify existing checks using `fullyAppliedLastArg`

lue-bird:
Many general checks/fixes were still taking the last arg as an argument instead of inferring it.

Through changing this, I also discovered a bug in `intersect` where `intersect empty` would be fixed to `empty` instead of `always empty`.

In a follow-up PR, I'd like to make a general fix/check cleanup.

### DIFF
--- a/tests/SimplifyTest.elm
+++ b/tests/SimplifyTest.elm
@@ -21526,7 +21526,9 @@ setIntersectTests =
             \() ->
                 """module A exposing (..)
 import Set
-a = Set.intersect set set
+a0 = Set.intersect
+a1 = Set.intersect set0
+a2 = Set.intersect set0 set1
 """
                     |> Review.Test.run ruleWithDefaults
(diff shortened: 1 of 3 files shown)


tests/SimplifyTest.elm
```diff
@@ -21526,7 +21526,9 @@ setIntersectTests =
             \() ->
                 """module A exposing (..)
 import Set
-a = Set.intersect set set
```
jfmengels, line 21529: `Set.insersect set set` should actually be reported, as it can be simplified to `set` (same for `Set.union`, and I imagine all the same things for `Dict`). I'll add them to #2.

lue-bird: Exactly. This is why I removed it from the "not reported" section. And thanks for adding it to the issue

jfmengels (approved):
Nice changes :+1:

PR #236, opened 2023-10-02
Make checkInfo argument implicit

jfmengels:
This is mostly to reduce the re-allocation of the functions that reference it. It also makes the function a little bit shorter.

I've only done it for a (small) part of the code because I'm not sure @lue-bird would be okay with it.
What do you think, is this a direction we should go into or not? If so, feel free to merge it, otherwise I'd be curious to discuss it.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -3359,9 +3359,9 @@ divisionChecks checkInfo =
 
 
 intDivideChecks : OperatorCheckInfo -> Maybe (Error {})
-intDivideChecks checkInfo =
+intDivideChecks =
     firstThatConstructsJust
-        [ \() ->
+        [ \checkInfo ->
             case AstHelpers.getUncomputedNumberValue checkInfo.right of
                 Just rightNumber ->
                     if rightNumber == 1 then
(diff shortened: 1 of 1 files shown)


lue-bird:
No, this was my plan all along but I was "worried" about currying overhead.
Do you have plans of e.g. changing the `-Checks` functions` into returning a list and applying the `firstThatConstructsJust` as late as possible?

jfmengels:
> No, this was my plan all along.

Nice :+1: 

> I find it much cleaner/atomic/readable but I was "worried" about currying overhead.

The currying overhead shouldn't matter too much here. It is slower, but because the functions are defined once at the start of the review application, and the only thing remaining is a function call with the last remaining argument, which is fast because it translates to a plain function call (without `A2/A3/...`).

> Do you have plans of e.g. changing the -Checks functions into returning a list and applying the firstThatConstructsJust as late as possible?

I hadn't thought of that. I'm not sure. I think it could make sense but if it puts the burden on the function that calls the check (and shouldn't necessary know whether there is one or more checks) then I'm not sure it's worth it. But feel free to try it out :+1: 

---

I'll merge this PR now so as not to create conflicts for other things, and let's do these changes as we go along and/or in other PRs like this.

PR #239, opened 2023-10-04
Qualified fn module generation

lue-bird:
- adds a codegen package fn modules generator
  - adds all package fn modules for `elm/`: `core`, `html`, `parser`, `random`, `json`. If we in the future want to support `elm/svg` or `elm-explorations/test` for example, this will take a manual step of `elm-codegen run` on the downloaded `elm.json`
  - adds an elm-review rule to replace the manual tuples + imports
  - applies the necessary fixes

I've looked through the leftover tuples and none of them were e.g. misspelled fns, so the rule seems to have caught everything.

If you think anything could be improved, like adding doc comments for the declarations and/or module, leave a comment.


jfmengels:
The #237 branch had conflicts so I had to rebase both branches and force push onto your branch. So please pull the changes before you make new commits.

Issue #242, opened 2023-10-04
Add unnecessaryCompositionAfterEmptyCheck for more operations


lue-bird:
```elm
Task.map f << Task.fail --> Task.fail
Task.mapError f << Task.succeed --> Task.succeed
Task.andThen f << Task.fail --> Task.fail
Task.onError f << Task.succeed --> Task.succeed

Json.Decode.map f << Json.Decode.fail --> Json.Decode.fail
Json.Decode.andThen f << Json.Decode.fail --> Json.Decode.fail

Result.map f << Err --> Err
-- done Result.mapError f << Ok --> Ok
Result.andThen f << Err --> Err
```

PR #243, opened 2023-10-04
Add missing "unnecessary composition after empty" simplifications

lue-bird:
Fully implements https://github.com/jfmengels/elm-review-simplify/issues/242


jfmengels (approved):
❤️

PR #245, opened 2023-10-05
Add andThen composition after wrap simplifications

lue-bird:
Adds `andThen f << wrap --> f` composition simplifications.
```elm
Maybe.andThen f << Just --> f
Result.andThen f << Ok --> f
Json.Decode.andThen f << Json.Decode.succeed --> f
Random.andThen f << Random.constant --> f
List.concatMap f << List.singleton --> f
Task.andThen f << Task.succeed --> f
Task.onError f << Task.fail --> f
```
If you think there are `andThen` functions missing in this list, please comment.

Fixing the `List.map` on singleton fix in the next PR.

### DIFF
--- a/tests/SimplifyTest.elm
+++ b/tests/SimplifyTest.elm
@@ -24741,6 +24741,24 @@ import Task
 a = a |> f
 """
                         ]
+        , test "should replace Task.andThen f << Task.succeed by f" <|
+            \() ->
+                """module A exposing (..)
+import Task
+a = Task.andThen f << Task.succeed
+"""
+                    |> Review.Test.run ruleWithDefaults
+                    |> Review.Test.expectErrors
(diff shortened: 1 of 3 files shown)


tests/SimplifyTest.elm
```diff
@@ -18526,7 +18526,8 @@ maybeAndThenTests =
         [ test "should not report Maybe.andThen used with okay arguments" <|
             \() ->
                 """module A exposing (..)
-a = Maybe.andThen f x
+a0 = Maybe.andThen f x
+a1 = Maybe.andThen << Just -- f arg missing
```
jfmengels: I think in general I would prefer these kinds of "don't report error given specific conditions" to be in their own tests that expect no errors, I feel like that would be better. But no need to change it now

PR #247, opened 2023-10-05
Map checks preserve wrap style of non-branching wrapped

lue-bird:
Previously, "map on wrap call" fixes did not try to preserve the "style" of wrap:
```elm
a |> List.singleton |> List.map f |> g
--> [ a |> f ] |> g
```
We now keep the existing style _for non-branching cases_:
```elm
a |> List.singleton |> List.map f |> g
--> (a |> f) |> List.singleton |> g
```
while branching styles are still over-written by using `[ ... ]` for list singletons
```elm
(if c then List.singleton a else List.singleton b) |> List.map f |> g
--> [ (if c then a else b) |> f) ] |> g
```
Not sure I'm super happy with the results but it's a bit nicer than before for sure.

Do you think we should also try to recover something like "the commonly used wrap style in all branches" or is that overkill?

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -5250,36 +5250,86 @@ listMapChecks =
 
 listMapOnSingletonCheck : CheckInfo -> Maybe (Error {})
 listMapOnSingletonCheck checkInfo =
+    -- we do not re-use mapWrapChecks because that would fix e.g.
+    -- map f (if c then List.singleton a else [ b ]) --> map f (List.singleton (if c then a else b))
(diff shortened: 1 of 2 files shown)


jfmengels (approved):
Thank you for looking into this :pray: 

> Do you think we should also try to recover something like "the commonly used wrap style in all branches" or is that overkill?

I think it's fine as it is. We can't always promise perfect fixes, but that's also partially why fixes are interactive, so that people can say no and resolve the issue themselves.

Issue #28, opened 2022-07-16
Reproduce the available simplifications we have for Set for Dict


jfmengels:
We have these checks for `Set` but we don't have them for `Dict`.

Help is welcome for tackling these! These don't all have to be done at the same time, so if you only want to tackle a few, that's absolutely fine. These are mostly gathered so we don't have dozens of issues.

```elm
Dict.map fn Dict.empty -- same for Dict.filter, Dict.remove...
--> Dict.empty

Dict.map (\k v -> v) list
--> list

Dict.map (\k v -> v)
--> identity

Dict.isEmpty Dict.empty
--> True

Dict.member x Dict.empty
--> False

Dict.fromList []
--> Dict.empty

Dict.toList Dict.empty
--> []

Dict.length Dict.empty
--> 0

Dict.intersect Dict.empty dict
--> Dict.empty

Dict.diff Dict.empty dict
--> Dict.empty

Dict.diff dict Dict.empty
--> dict

Dict.union dict Dict.empty
--> dict

Dict.insert x y Dict.empty
--> Dict.singleton x y

Dict.partition fn Dict.empty
--> ( Dict.empty, Dict.empty )

Dict.partition ((\k v -> True) dict
--> ( list, Dict.empty )

Dict.partition ((\k v -> False) dict
--> ( Dict.empty, list )
```

lue-bird:
Ok so the only one left now is
```elm
Dict.insert x y Dict.empty
--> Dict.singleton x y
```
which I'm not super fond of in pipelines
```elm
Dict.empty
    |> Dict.insert x y
    |> Dict.insert z w
-->
Dict.singleton x y
    |> Dict.insert z w
```
I guess there could be an argument made (just like with cons) that this then should be converted into
```elm
Dict.fromList [ ( x, y ), ( z, w ) ]
```
but this to me feels again like a code-style change.

jfmengels:
> Ok so the only one left now is

There's also the fold functions which we don't have for `Dict`.

> but this to me feels again like a code-style change.

Yeah, it's always a bit on the edge. I *think* it would be fine, but... :man_shrugging: 
If you want to do it, let's do it. Otherwise we could leave it alone. I'm not sure either at this point :thinking:

PR #248, opened 2023-10-05
Add Dict.map simplifications

lue-bird:
Part of #28 
```elm
Dict.map f Dict.empty -- same for Dict.filter, Dict.remove...
--> Dict.empty

Dict.map (\_ v -> v) dict
--> dict
```


jfmengels:
:heart_eyes:

PR #249, opened 2023-10-05
Add script to run code generation

jfmengels:
- Add the missing `elm-codegen` dependency
- Add script to run code generation automatically
- Fix generation issues due to type variants and function names conflicting. Now if the module exposing `problem` and `Problem`, then we'll respectively generate `problem` and `problemVariant`
- Re-generate the Fn modules to include the variants

Please try it out using `npm run codegen`

### DIFF
--- a/codegen/Generate.elm
+++ b/codegen/Generate.elm
@@ -38,18 +38,14 @@ moduleDocsToFile moduleDocs =
         moduleName : List String
         moduleName =
             moduleDocs.name |> String.split "."
-
-        values : List String
-        values =
-            List.map .name moduleDocs.values
     in
     Elm.file ("Fn" :: moduleName)
-        (List.map (\name -> declaration moduleName ( name, name )) values
+        (List.map (\{ name } -> declaration moduleName ( name, name )) moduleDocs.values
             ++ (moduleDocs.unions
                     |> List.concatMap .tags
                     |> List.map
                         (\( tagName, _ ) ->
-                            declaration moduleName ( tagDeclarationName values (stringFirstCharToLower tagName), tagName )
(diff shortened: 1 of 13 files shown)


codegen/Generate.elm
```diff
@@ -38,29 +38,43 @@ moduleDocsToFile moduleDocs =
...
+    Elm.declaration declarationName
+        (Elm.tuple
+            (Elm.list (moduleName |> List.map Elm.string)
+                |> Elm.withType Gen.Elm.Syntax.ModuleName.annotation_.moduleName
+            )
+            (Elm.string originalName)
+        )
+
+
+tagDeclarationName : List String -> String -> String
```
lue-bird, line 69: I'd be fine with suffixing _every_ variant or even creating e.g. `Fn.Result.Variant.err`

lue-bird:
I think the only thing left to change is to add the same suffixing check to the elm-review rule

codegen/Generate.elm
jfmengels: Yeah, that's probably a good idea, less confusing :+1:

jfmengels: We'll end up with things like `Fn.Maybe.justVariant` and `Fn.Basics.trueVariant`. Does that sound okay to you? I'm fine with it, but it is definitely not as nice for the common case. In fact, I don't know if we'll ever run into a case where we'll need to refer to a variant that would otherwise result in a conflict (at least today we don't).

lue-bird: sounds good I think.

jfmengels: Alright, let's try that then. We can always undo this later otherwise.

lue-bird (approved):

PR #250, opened 2023-10-05
Add Dict.filter simplifications

lue-bird:
As hinted at in #28 
```elm
Dict.filter f Dict.empty
--> Dict.empty

Dict.filter (\_ _ -> True) dict
--> dict

Dict.filter (\_ _ -> False) dict
--> Dict.empty
```
Bonus: Correct `Evaluate.isAlwaysBool` (terrible name and seems unnecessary anyway?), add `Set.filter` simplifications to summary.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -951,6 +951,15 @@ Destructuring using case expressions
     Dict.member x Dict.empty
     --> False
 
+    Dict.filter f Dict.empty
+    --> Dict.empty
+
+    Dict.filter (\_ _ -> True) dict
+    --> dict
+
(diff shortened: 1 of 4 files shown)


src/Simplify.elm
```diff
@@ -917,6 +917,15 @@ Destructuring using case expressions
     List.foldl f x (Set.toList set)
     --> Set.foldl f x set
 
+    Set.filter f Set.empty
+    --> Set.empty
+
+    Set.filter (\_ -> True) set
+    --> dict
```
jfmengels, line 924:
```suggestion
    --> set
```

PR #253, opened 2023-10-05
Correct Dict.partition simplifications

lue-bird:
Someone (probably me) thought that the function given to `Dict.partition` does not take a key.
By pure chance, a previous bug in `Evaluate.isAlwaysBoolean` made this fix still work correctly (mostly).

With this PR, Dict.partition gets its own checks independent from `collectionPartitionChecks` + correct checks.


jfmengels (approved):
Good catch!

PR #254, opened 2023-10-06
Add Dict.foldlr simplifications and make existing Dict functions with extra arg generic

lue-bird:
`Dict` simplifications should now be on par with `Set` except for `insert`
```elm
-- same for foldr
Dict.foldl f initial Dict.empty
--> initial

Dict.foldl (\_ soFar -> soFar) initial dict
--> initial
```

Bonus: Convert existing manual Dict operation checks that exist because the function takes 2 args to more generic helpers.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -10162,7 +10162,7 @@ partitionOnEmptyChecks emptiable =
 
 partitionWithConstantFunctionResult : Node Expression -> TypeProperties (EmptiableProperties ConstantProperties otherProperties) -> CheckInfo -> Maybe (Error {})
 partitionWithConstantFunctionResult constantFunctionResult collection checkInfo =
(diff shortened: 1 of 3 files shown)


src/Simplify.elm
```diff
@@ -8783,6 +8765,15 @@ subCollection =
     }
 
 
+{-| The map checks
+
+    map f empty --> empty
+
+    map identity emptiable --> emptiable
+
+If your mapping function also takes extra information like the key or index as an argument, use `emptiableMapWithExtraArgChecks`.
```
jfmengels, line 8679: Just wanted to say that I think that these documentation comments are really valuable, this is great :heart:

jfmengels (commented):
Great work! There is just one tiny nit remaining, let me know if I misunderstood something about that.

Issue #127, opened 2023-08-15
Add +, * simplifications for List.sum and List.product


lue-bird:
when `expectNaN` is disabled
```elm
List.sum [ a, n, b, c - n ]
--> List.sum [ a, b, c ]

List.sum [ a, n, b, c + (-n) ]
--> List.sum [ a, b, c ]

List.product [ a, 0, b ]
--> 0
```
always
```elm
List.sum [ a, 0, b ]
--> List.sum [ a, b ]

List.product [ a, 1, b ]
--> List.product [ a, b ]
```

jfmengels:
To add to the list:

When `expectNaN` is disabled
```elm
List.sum [ a, n, b, - n ]
--> List.sum [ a, b ]

List.product [ a, 0, b ]
--> 0
```

lue-bird:
```elm
List.product [ a, 0, b ]
--> 0
```
was already in the list.
```elm
List.sum [ a, n, b, -n ]
--> List.sum [ a, b ]
```
in my mind was covered by
```elm
List.sum [ a, n, b, c + (-n) ]
--> List.sum [ a, b, c ]
```
but I added it to the PR description.

PR #256, opened 2023-10-06
List.sum on list with 0, List.product on list with 1

lue-bird:
Two simplifications from https://github.com/jfmengels/elm-review-simplify/issues/127
```elm
List.sum [ a, 0, b ]
--> List.sum [ a, b ]

List.product [ a, 1, b ]
--> List.product [ a, b ]
```

### DIFF
--- a/src/Simplify/Evaluate.elm
+++ b/src/Simplify/Evaluate.elm
@@ -1,4 +1,4 @@
-module Simplify.Evaluate exposing (getBoolean, getInt, isAlwaysBoolean, isEqualToSomethingFunction)
+module Simplify.Evaluate exposing (getBoolean, getInt, getNumber, isAlwaysBoolean, isEqualToSomethingFunction)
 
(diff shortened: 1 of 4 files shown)


src/Simplify/Evaluate.elm
```diff
@@ -153,3 +153,31 @@ getInt resources baseNode =
...
+
+
+getNumber : Infer.Resources a -> Node Expression -> Maybe Float
+getNumber resources baseNode =
+    let
+        unparenthesized : Node Expression
+        unparenthesized =
+            AstHelpers.removeParens baseNode
+    in
+    case getInt resources unparenthesized of
```
jfmengels, line 165: We should ideally support floats as well, so that `List.sum [ a, 0.0, b ]` also gets simplified.

jfmengels (approved):
Very nice :blush:

PR #257, opened 2023-10-07
operations on list containing absorbing element

lue-bird:
#### when `expectNaN` is not enabled
```elm
List.product [ a, 0, b ]
--> 0
```
which is a simplification from https://github.com/jfmengels/elm-review-simplify/issues/127
#### when `expectNaN` is enabled
```elm
List.sum [ a, 0 / 0, b ]
--> 0 / 0

List.product [ a, 0 / 0, b ]
--> 0 / 0
```


jfmengels:
Very nice simplifications!

PR #261, opened 2023-10-07
Add List.all/any with not on list with (not absorbing) simplifications

lue-bird:
as suggested in https://github.com/jfmengels/elm-review-simplify/pull/260#discussion_r1349517443
```elm
List.all not [ a, True, b ]
--> False

List.any not [ a, False, b ]
--> True
```
Bonus: Improved error for `any/ all identity [ .. absorbing .. ]`

### DIFF
--- a/tests/SimplifyTest.elm
+++ b/tests/SimplifyTest.elm
@@ -13352,14 +13352,14 @@ a = List.all (always True)
 a = always True
 """
                         ]
-        , test "should not report List.all on list with False but non-identity function" <|
+        , test "should not report List.all on list with False but non-identity and non-not function" <|
             \() ->
                 """module A exposing (..)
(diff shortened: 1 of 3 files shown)


tests/SimplifyTest.elm
```diff
@@ -13380,6 +13380,29 @@ a = List.all identity [ b, False, c ]
                             }
                             |> Review.Test.whenFixed """module A exposing (..)
 a = False
+"""
+                        ]
+        , test "should not report List.all not on list with False and no True" <|
+            \() ->
+                """module A exposing (..)
+a = List.all not [ b, False, c ]
```
jfmengels, line 13388: We could replace this by `List.all not [ b, c ]`

jfmengels (approved):

PR #263, opened 2023-10-08
Add simplifications for number conversions

### LINKED ISSUE #259: Basic operations on integer literals
In the following examples I'm using `1` but imagine any literal integer (so `1` but not `1.0`)

- [ ] `toFloat 1` -> `1`
- [ ] `round 1` -> `1`
- [ ] `ceiling 1` -> `1`
- [ ] `floor 1` -> `1`
- [ ] `truncate 1` -> `1`

Compositions of these functions:
- [ ] `round << toFloat` -> `identity`
- [ ] `ceiling << toFloat` -> `identity`
- [ ] `floor << toFloat` -> `identity`
- [ ] `truncate << toFloat` -> `identity`

jfmengels:
Fixes #259

- `toFloat 1` -> `1`
- `round 1` -> `1`
- `ceiling 1` -> `1`
- `floor 1` -> `1`
- `truncate 1` -> `1`

Compositions of these functions:
- `round << toFloat` -> `identity`
- `ceiling << toFloat` -> `identity`
- `floor << toFloat` -> `identity`
- `truncate << toFloat` -> `identity`

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -4158,13 +4158,13 @@ basicsToFloatChecks checkInfo =
 intToIntChecks : CheckInfo -> Maybe (Error {})
 intToIntChecks =
     firstThatConstructsJust
-        [ intToIntCheck
+        [ unnecessaryConversionToIntOnIntCheck
         , onCallToInverseReturnsItsArgumentCheck Fn.Basics.toFloat
         ]
 
 
-intToIntCheck : CheckInfo -> Maybe (Error {})
-intToIntCheck checkInfo =
+unnecessaryConversionToIntOnIntCheck : CheckInfo -> Maybe (Error {})
+unnecessaryConversionToIntOnIntCheck checkInfo =
     case Evaluate.getInt checkInfo checkInfo.firstArg of
         Just _ ->
             Just
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -200,6 +200,12 @@ Destructuring using case expressions
     toFloat 1
     --> 1
 
+    round 1
+    --> 1
+
+    round (toFloat n)
+    --> n
+
 
 ### Lambdas
 
@@ -2723,6 +2729,7 @@ functionCallChecks =
         , ( Fn.Basics.not, ( 1, basicsNotChecks ) )
         , ( Fn.Basics.negate, ( 1, basicsNegateChecks ) )
         , ( Fn.Basics.toFloat, ( 1, basicsToFloatChecks ) )
+        , ( Fn.Basics.round, ( 1, basicsRoundChecks ) )
         , ( Fn.Tuple.first, ( 1, tupleFirstChecks ) )
         , ( Fn.Tuple.second, ( 1, tupleSecondChecks ) )
         , ( Fn.Tuple.pair, ( 2, tuplePairChecks ) )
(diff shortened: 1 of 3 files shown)


src/Simplify.elm
```diff
@@ -4108,6 +4135,54 @@ isNegatableOperator op =
...
 
+basicsToFloatChecks : CheckInfo -> Maybe (Error {})
+basicsToFloatChecks checkInfo =
+    case Evaluate.getNumber checkInfo checkInfo.firstArg of
+        Just _ ->
+            Just
+                (Rule.errorWithFix
+                    { message = "Unnecessary " ++ qualifiedToString (qualify checkInfo.fn defaultQualifyResources) ++ " on a literal number"
+                    , details =
+                        [ "A literal integers is considered as both an Int and a Float, there is therefore no need to explicitly convert it to a Float."
```
lue-bird, line 4146: Typo "integers" → "integer" and maybe mention `number`? Something like

"A literal integer is considered a `number` which means it can be used as both an Int and a Float and there is no need to explicitly convert it to a Float."

src/Simplify.elm
```diff
@@ -4108,6 +4135,54 @@ isNegatableOperator op =
             Nothing
 
 
+basicsToFloatChecks : CheckInfo -> Maybe (Error {})
+basicsToFloatChecks checkInfo =
+    case Evaluate.getNumber checkInfo checkInfo.firstArg of
```
lue-bird, line 4140: You'll have to use `getInt` because `getNumber` also parses `1.2` and inferred as `1.2` for example.
If you think the naming is off and should be `getFloat`, feel free to change it (along with AstHelpers.getUncomputedNumber).

Tests for this would also be nice.

src/Simplify.elm
```diff
@@ -4108,6 +4135,54 @@ isNegatableOperator op =
...
+
+intToIntCheck : CheckInfo -> Maybe (Error {})
+intToIntCheck checkInfo =
+    case Evaluate.getInt checkInfo checkInfo.firstArg of
+        Just _ ->
+            Just
+                (Rule.errorWithFix
+                    { message = "Unnecessary integer conversion on a literal integer"
+                    , details =
+                        [ "Literal integers are already considered to be integers and it is therefore not necessary to convert them further."
```
lue-bird, line 4174: Maybe singular and say Int instead of integer? Something like

"A literal integer is already considered to be an Int which means converting it further is unnecessary."

src/Simplify.elm
```diff
@@ -4108,6 +4135,54 @@ isNegatableOperator op =
...
+
+intToIntChecks : CheckInfo -> Maybe (Error {})
+intToIntChecks =
+    firstThatConstructsJust
+        [ intToIntCheck
+        , onCallToInverseReturnsItsArgumentCheck Fn.Basics.toFloat
+        ]
+
+
+intToIntCheck : CheckInfo -> Maybe (Error {})
```
lue-bird, line 4166: Since the name `intToIntChecks` is already used, maybe call it something like `unnecessaryConversionToIntOnIntCheck`

src/Simplify.elm
jfmengels: Nah, the naming is fine. I copy-pasted from somewhere thinking I should change it to `getInt` and then somehow forgot. I'll add a test.

jfmengels:
All great suggestions, thank you :+1:

lue-bird (approved):

PR #264, opened 2023-10-08
Rework Indexable, Collection, FromList properties

lue-bird:
I'll apologize in advance for the first commit which one could say does multiple things at once. Sorry!
- Introduce property `elements.get` which can detect known elements, including information of whether some parts are unknown
  - allowed introducing generic checks for e.g. member
  - `onCollectionWithAbsorbingCheck` now also reports e.g. `List.all identity (a :: False :: bs)`, `callOnListWithAbsorbingElement` therefore became unnecessary
- `Indexable` was confusing. Does `literalElements` refer to the "fromList" construction method? I would expect no but some checks were relying on that behavior.
  - → remove in favor of `fromList.getList` and `elements.get`
- `FromListProperties` contained `unionLeftElementsStayOnTheLeft`. To me it does not seem like the right place. First, not every type that can be constructed from a list has a union function. Second, a type can have different union functions, e.g. `Array.append`, the pipe-friendly version `prependTo` and the pipe-friendly prepend function `appendTo`.
  - `leftElementsStayOnTheLeft` is now provided as an argument to the union checks
- Simplify `FromListProperties` to match any `fromList` call, not just on literals
- Refactor `-DetermineSize` functions
- Introduce separate declarations for the grouped properties of a type

(Unrelated note: I've added my personal todo list for simplify publically in https://github.com/lue-bird/elm-review-simplify/issues/1 which always contains all the refactorings I plan to do in the near future but have not yet formalized)

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -9271,9 +9271,9 @@ dictDetermineSize resources =
                     case listGetElements resources fromListCall.firstArg of
                         Just listElements ->
                             if listElements.allKnown then
-                                case traverse getComparableExpressionInTupleFirst listElements.known of
-                                    Just comparableExpressions ->
-                                        comparableExpressions |> unique |> List.length |> Exactly |> Just
+                                case traverse (getTupleWithComparableFirst resources.lookupTable) listElements.known of
+                                    Just comparableKeyExpressions ->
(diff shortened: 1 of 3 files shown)


jfmengels:
All very nice changes :+1:

PR #265, opened 2023-10-09
Add missing List.take .drop on [] to simplification list

morteako:
This adds 

```
List.take n []
--> []
List.drop n []
--> []
```

to the documentation list.
I could not find them there, and I think they should be included (if i understand correctly).
But it is possible they are included by some general "on empty lists"-statement?

These are currently implemented, executed and in the test suite:
https://github.com/jfmengels/elm-review-simplify/blob/1180d1c1a795e645759b1d4abe12b565a4a686da/tests/SimplifyTest.elm#L15227

https://github.com/jfmengels/elm-review-simplify/blob/1180d1c1a795e645759b1d4abe12b565a4a686da/tests/SimplifyTest.elm#L15305

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -783,6 +783,12 @@ Destructuring using case expressions
     List.drop 0 list
     --> list
 
+    List.take n []
+    --> []
+
+    List.drop n []
+    --> []
+
     List.reverse []
     --> []
 


lue-bird:
We have this line I guess: https://github.com/jfmengels/elm-review-simplify/pull/265/files#diff-6239bccafe2b75c7e560542732acf25d142d08608e1e0f0ffe9b53c09cfb1e0eR566

but some functions like intersperse have the `List.operation ... [] --> []` listed regardless which is inconsistent.

I'd prefer removing those instead

morteako:
> We have this line I guess: https://github.com/jfmengels/elm-review-simplify/pull/265/files#diff-6239bccafe2b75c7e560542732acf25d142d08608e1e0f0ffe9b53c09cfb1e0eR566
> 
> but some functions like intersperse have the `List.operation ... [] --> []` listed regardless which is inconsistent.
> 
> I'd prefer removing those instead.

Ah, yes. My bad.

I guess it makes it a bit harder to search for simplifications for different functions,
but that is probably not done very often by most people using elm-review-simplify.

PR #269, opened 2023-10-09
Add access record type alias constructor call simplification

lue-bird:
Part of https://github.com/jfmengels/elm-review-simplify/issues/162
```elm
(Record first second).first --> first
```
with record type aliases defined in the current module, project or a dependency.

Something really strange I've noticed: The changes from https://github.com/jfmengels/elm-review-simplify/pull/206 and https://github.com/jfmengels/elm-review-simplify/pull/210/ seem to be completely lost (maybe even more?). Maybe these we're force-overwritten or something (https://github.com/jfmengels/elm-review-simplify/pull/207 especially seems sus). At any rate, I'll try to re-apply these changes in the next PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,7 @@
 - `floor 1` to `1`
 - `truncate 1` to `1`
 - `round (toFloat n)` to `n` (and same for ceiling, floor, truncate)
+- `(Record first second).first` to `first`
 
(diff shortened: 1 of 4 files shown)


jfmengels:
Thanks, this is a nice change :blush: 

I'll look into recovering the lost PRs, I imagine that that's my fault :see_no_evil:

Issue #268, opened 2023-10-09
More simplifications for List.take and List.drop


morteako:
# More simplifications for List.take and List.drop

Some suggestions for List.take and List.drop simplifications.
I think take and map are functions that are used a lot, so new rules would probably be applied in many codebases 🚀  

(`n` is refering to the Int argument , as in `List.take n list`)

### List.drop with negative n
```elm
List.drop -1 list
--> list
```
### List.length on List.take with non-negative n
```elm
List.length (List.take 10 list)
--> 10
```

### Literal lists with size <= n
```elm
List.take 5 [a,b,c]
--> [a,b,c]

List.drop 5 [a,b,c]
--> []

List.take 1 (List.singleton x)
--> List.singleton x

List.drop 1 (List.singleton x)
--> []
```

### Take on take, drop on drop
```elm
list
|> List.take n
|> List.take m
--> list |> List.take (min n m)

list
|> List.drop n
|> List.drop m
--> list |> List.drop (n + m)
```

### List.repeat with non-negative n <= repeat arg
```elm
List.take 3 (List.repeat 5 x )
--> List.repeat 2 x

List.drop 3 (List.repeat 5 x )
--> List.drop 2 x
```

### List.range with size >= n
```elm
List.take 3 (List.range 0 10)
--> List.range 0 2

List.drop 3 (List.range 0 10)
--> List.range 3 10
```

### Optimizations when combining with List.map/filterMap
```elm
-- Same for List.drop and same for List.filterMap
list
|> List.map f
|> List.take n
--> 
list
|> List.take n
|> List.map f
```

---
### Generalization question

Some of these can be generalized more, f.ex.

```elm
-- n > 0
List.drop n (List.repeat m x )
--> List.repeat (m-n) x
```
But I am not sure about the clarity of the resulting code and also about how it will impact further simplifications.
So I went with the simpler one. Open for suggestions

---
Feedback and tips welcome! And of course more suggestions and corrections.

I can probably look into implementing most of these myself.

lue-bird:
Nice.
- ```elm
  List.length (List.take 10 list)
  --> 10
  ```
  doesn't work because list can have less than 10 elements.

- The take and drop ones are already tracked in https://github.com/jfmengels/elm-review-simplify/issues/2

- ```elm
  List.take 3 (List.repeat 5 x)
  --> List.repeat 2 x
  ```
  should be repeat 3

Other than those nitpicks, looks good

morteako:
[...]

PR #270, opened 2023-10-10
More List.drop simplifications

lue-bird:
Some simplifications from https://github.com/jfmengels/elm-review-simplify/issues/268 and more
```elm
List.drop -1 list
--> list

List.drop 3 [ a, b ] -- including List.singleton etc where we can determine the size
--> []

List.drop 2 [ a, b, c ]
--> [ c ]
```

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -786,6 +786,12 @@ Destructuring using case expressions
     List.drop 0 list
     --> list
 
+    List.drop 3 [ a, b ]
+    --> []
+
+    List.drop 2 [ a, b, c ]
+    --> [ c ]
+
     List.reverse []
     --> []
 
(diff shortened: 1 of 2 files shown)


tests/Simplify/ListTest.elm
```diff
@@ -7118,7 +7118,7 @@ a = List.drop 0 list
                     |> Review.Test.run ruleWithDefaults
                     |> Review.Test.expectErrors
                         [ Review.Test.error
-                            { message = "List.drop 0 will always return the same given list"
+                            { message = "List.drop with count 0 will always return the same given list"
```
jfmengels, line 7121: I feel like the previous error message was better :man_shrugging:

jfmengels (approved):
Nice. I'm unsure about the count error message, but let's go with that if you think it's better.

src/Simplify.elm
```diff
@@ -7138,19 +7157,102 @@ listDropChecks : CheckInfo -> Maybe (Error {})
...
+        ]
+
+
+{-| The drop check
+
+    drop n (collection with size <= n) --> empty
+
+So for example
+
+    Array.drop 3 (Array.repeat 2 a) --> []
```
morteako, line 7187: I think Array does not have a drop-function, only slice.

lue-bird: Thanks!

PR #273, opened 2023-10-10
WIP: cleanup (again)

lue-bird:
- [x] Introduce more general checks for sequence, join, and repeatFlat
- [x] Introduce `MappableProperties`
- [x] Use "can be combined checks" more
- [x] Organize helpers better into sections
- [x] Introduce helpers to create type subset properties

### to discuss
- I'd like to rename "andThen" checks to "mapFlat" which is in my mind easier to understand as `concatMap` and has a nice conststency with e.g. "repeatFlat" which is effectively `repeat >> andThen identity`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,11 @@
 - `Array.toList Array.empty` to `[]`
 - `Array.toList (Array.repeat n a)` to `List.repeat n a`
 - `Array.toIndexedList Array.empty` to `[]`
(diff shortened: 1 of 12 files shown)


jfmengels:
You have some conflicts because of my restoration of old PR (sorry :see_no_evil:). Let me know if you want some help with that.

> I'd like to rename "andThen" checks to "mapFlat"

Go ahead. These names are all quite abstract at this point anyway (but not Haskell-like abstract).

PR #277, opened 2023-10-10
Cleanup

lue-bird:
Continuing #273 which had problems due to conflicts that I did not not know how to fix in the non-renamed remote branch of the PR.

All changes old + new:

- Introduce more general checks for sequence, join, repeatFlat and random checks
- Introduce `MappableProperties`
- Use "can be combined checks" more
- Organize helpers better into sections
- Introduce helpers to create type subset properties
- Fixed an issue where `Random.map << always` would be fixed to `Random.constant`
- Fixed an issue where `List.any (\x -> x == (f x)) list` would be fixed to `List.member (f x) list`
- Add documentation to checks with generic types

Can be merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,7 +103,7 @@ Bug fixes:
 - Fixed an issue where `Dict.partition (always (always True/False)) dict` would not be reported
 - Fixed an issue where `List.filterMap f [ a, Nothing, b ]` would be fixed to `List.filterMap f [ a, b ]`
 - Fixed an issue where `Random.map << always` would be fixed to `Random.constant`
-- Fixed an issue where `List.any (\\z -> x == y) list` would be fixed to `List.member (f x) list`
+- Fixed an issue where `List.any (\x -> x == (f x)) list` would be fixed to `List.member (f x) list`
 
 ## [2.1.2] - 2023-09-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,7 @@ Bug fixes:
 - Fixed an issue where `Dict.partition (always (always True/False)) dict` would not be reported
 - Fixed an issue where `List.filterMap f [ a, Nothing, b ]` would be fixed to `List.filterMap f [ a, b ]`
 - Fixed an issue where `Random.map << always` would be fixed to `Random.constant`
+- Fixed an issue where `List.any (\\z -> x == y) list` would be fixed to `List.member (f x) list`
 
 ## [2.1.2] - 2023-09-28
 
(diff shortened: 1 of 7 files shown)


lue-bird:
Can be reviewed/merged. I won't add anything from now.

src/Simplify/AstHelpers.elm
```diff
@@ -95,6 +101,98 @@ removeParensFromPattern patternNode =
...
+subExpressions : Expression -> List (Node Expression)
+subExpressions expression =
+    case expression of
+        Expression.LetExpression letBlock ->
+            letBlock.expression
+                :: (letBlock.declarations
+                        |> List.map Node.value
+                        |> List.map
+                            (\letDeclaration ->
+                                case letDeclaration of
```
jfmengels, line 115:
```suggestion
                :: (letBlock.declarations
                        |> List.map
                            (\letDeclaration ->
                                case Node.value letDeclaration of
```

CHANGELOG.md
```diff
@@ -103,6 +103,7 @@ Bug fixes:
 - Fixed an issue where `Dict.partition (always (always True/False)) dict` would not be reported
 - Fixed an issue where `List.filterMap f [ a, Nothing, b ]` would be fixed to `List.filterMap f [ a, b ]`
 - Fixed an issue where `Random.map << always` would be fixed to `Random.constant`
+- Fixed an issue where `List.any (\\z -> x == y) list` would be fixed to `List.member (f x) list`
```
jfmengels: Where did the `f` come from in this example?

src/Simplify/AstHelpers.elm
lue-bird: Right. Fixed

CHANGELOG.md
lue-bird: Corrected in one of the next commits

src/Simplify.elm
```diff
@@ -8777,6 +8777,17 @@ isEqualToSomethingFunction rawNode =
...
 
+expressionContainsVariable : String -> Expression -> Bool
+expressionContainsVariable variableToFind expressionNode =
+    case expressionNode of
+        Expression.FunctionOrValue [] variable ->
+            variable == variableToFind
+
+        nonVariableExpressionNode ->
+            List.any (\(Node _ sub) -> expressionContainsVariable variableToFind sub)
+                (AstHelpers.subExpressions nonVariableExpressionNode)
```
jfmengels, line 8858: This sounds like a pretty expensive function, because this is (unlikely but) potentially a large expression, and we'll be going through every node of it. But I don't have a better solution except abandoning this check when it looks to be this complex.
(So let's leave it as is :+1: )

jfmengels (approved):
Nice work!

I see you have moved a lot of the checks into sections. Do you think we should move them into separate files? Or have a file with generic checks? I guess the biggest gain would be discoverability of the checks (what's exported makes that clear), but not entirely sure it's worth it, especially as that would mean a lot of things would need to be split into new files.

PR #278, opened 2023-10-13
Add record type alias constructor simplifications

lue-bird:
Everything left in https://github.com/jfmengels/elm-review-simplify/issues/162  and more
```elm
-- given
type alias Record = { first : Int, second : Int }

-- all those that exist for literal records, for example
(if c then Record first second else Record second first).first
--> if c then (Record first second).first else (Record second first).first

.first << Record first
--> always first

.second << Record first
--> identity
```


jfmengels (approved):
:raised_hands:

Issue #246, opened 2023-10-05
Automate composition simplifications


jfmengels:
In PRs like #245, we specify composition checks that we already do in plain function calls.

I wonder if we could have a way to specify compositions in one place and then automatically derive the same simplification for plain function calls, so that we get two simplifications for the price of one.

I haven't put in much thought about a system like this yet, but the thought has crossed my mind several times, so I wanted to put it in writing so that we could potentially start designing around this.

PR #279, opened 2023-10-13
Combined call and composition checks

lue-bird:
Refactor based on the idea in #246 
- Convert all checks into combined "`IntoFnCheck`"s
- Introduce module-level `IntoFnCheck` declarations for each fn
- Go through each "use together / in combination with `...` / for the same thing as a ... check" and convert to a `IntoFnCheck`

Your idea works really well in making checks simpler and more reliable. Props for coming up with it!

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -4966,11 +4966,20 @@ listMapChecks =
     intoFnChecksFirstThatConstructsError
         [ emptiableMapChecks listCollection
         , listMapOnSingletonCheck
-        , dictToListMapChecks
         , onSpecificFnCallCanBeCombinedCheck
             { args = [ tupleSecondAccessFunctionProperties ]
             , earlierFn = Fn.Array.toIndexedList
             , combinedFn = Fn.Array.toList
+            }
+        , onSpecificFnCallCanBeCombinedCheck
+            { args = [ tupleFirstAccessFunctionProperties ]
+            , earlierFn = Fn.Dict.toList
+            , combinedFn = Fn.Dict.keys
+            }
+        , onSpecificFnCallCanBeCombinedCheck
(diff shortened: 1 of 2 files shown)


jfmengels:
> If rebasing means I'll have to reopen a new PR, so be it. This PR is mostly to track progress for me but more transparently.

You're good to go :+1: :wink: 

And I'm fine with you making multiple branches/PRs at the same time, even if they were to conflict (we can always solve them, no worries).

lue-bird:
This should be it.
Comment if you notice more checks where would could combine composition+call or if some edge case should be tested.

src/Simplify.elm
```diff
@@ -10115,61 +10105,73 @@ Examples:
...
+                        ++ " with "
+                        ++ firstArg.description
+                        ++ String.concat (List.map (\arg -> " and " ++ arg.description) argsAfterFirst)
+
+        laterInitArgsAreValidForOperation : List (Node Expression) -> Infer.Resources res -> Bool
+        laterInitArgsAreValidForOperation laterInitArgs resources =
+            List.all identity
+                (List.map2 (\arg argConfig -> argConfig.is (extractInferResources resources) arg)
+                    laterInitArgs
+                    config.args
```
jfmengels, line 10049: Could there be a problem here if the length of the 2 lists is not the same? :thinking:

lue-bird: That's why we additionally check for `fullyAppliedLastArg` and `onlyLastArgIsCurried checkInfo.later` in the call and composition cases.

jfmengels (approved):
Really nice! You made some nice new architectures that seem like they work quite well :blush: 

I haven't looked if you missed some compositions but I'm sure we'll find them over time.

Great work :blush:

PR #281, opened 2023-10-15
Replace union, intersect with the same args by the last arg

jfmengels:
Implements simplifications from #2:

```elm
Dict.union dict dict
--> dict

Dict.intersect dict dict
--> dict

Set.union set set
--> set

Set.intersect set set
--> set
```

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1006,6 +1006,9 @@ Destructuring using case expressions
     Set.union set Set.empty
     --> set
 
+    Set.union set set
+    --> set
+
     Set.union (Set.fromList [ a, b ]) (Set.fromList [ c, d ])
     --> Set.fromList [ a, b, c, d]
 
@@ -5818,7 +5821,10 @@ setDiffChecks =
 
 setUnionChecks : IntoFnCheck
 setUnionChecks =
-    intoFnCheckOnlyCall (collectionUnionChecks { leftElementsStayOnTheLeft = True } setCollection)
+    intoFnChecksFirstThatConstructsError
+        [ intoFnCheckOnlyCall (collectionUnionChecks { leftElementsStayOnTheLeft = True } setCollection)
+        , whenArgumentsAreEqualReturnLastCheck
+        ]
 
 
 setMapChecks : IntoFnCheck
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1089,6 +1089,9 @@ Destructuring using case expressions
     Dict.union dict Dict.empty
     --> dict
 
+    Dict.union dict dict
+    --> dict
+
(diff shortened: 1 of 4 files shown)


src/Simplify.elm
```diff
@@ -5887,7 +5905,10 @@ dictMapChecks =
 
 dictIntersectChecks : IntoFnCheck
 dictIntersectChecks =
-    collectionIntersectChecks dictCollection
+    intoFnChecksFirstThatConstructsError
+        [ collectionIntersectChecks dictCollection
+        , whenArgumentsAreEqualReturnLastCheck
```
lue-bird, line 5910: Why not include `whenArgumentsAreEqualReturnLastCheck` directly in `collectionIntersectChecks`?

lue-bird (approved):

src/Simplify.elm
```diff
@@ -10425,6 +10449,43 @@ operationDoesNotChangeResultOfOperationCheck =
...
+Examples:
+
+    Dict.union dict dict
+    --> dict
+
+    Set.union set set
+    --> set
+
+-}
+whenArgumentsAreEqualReturnLastCheck : IntoFnCheck
```
lue-bird, line 10466: Maybe make clear in the name that this only applied two the first 2 arguments. Something like `onTwoEqualArgumentsReturnsLast`

src/Simplify.elm
```diff
@@ -10425,6 +10449,43 @@ operationDoesNotChangeResultOfOperationCheck =
...
+whenArgumentsAreEqualReturnLastCheck =
+    intoFnCheckOnlyCall
+        (\checkInfo ->
+            case secondArg checkInfo of
+                Just secondArg_ ->
+                    case Normalize.compare checkInfo checkInfo.firstArg secondArg_ of
+                        Normalize.ConfirmedEquality ->
+                            Just
+                                (returnsArgError
+                                    (qualifiedToString checkInfo.fn ++ " where the first and second argument are equal")
```
lue-bird, line 10476: Maybe shorten to "with equal first and second arguments"

src/Simplify.elm
lue-bird: Any reason why you chose to name it `onEqualFirstArgumentsReturnLastCheck` (not "two" and not "with")?

jfmengels: `on` because that's what you suggested at first before editing (and I thought that `on` was kind of the new naming convention, not sure what's best :shrug: ). And I wanted to name it `onTwoFirst...` or `onFirstTwo...` but that sounded more confusing than helpful.

I'll name it `withTwoEqualArgumentsReturnsLast`. Should I suffix it with "check"?

lue-bird: Yeah, with -check

PR #283, opened 2023-10-16
Add ++ case for listGetElements

morteako:
Add case for `++` in `listGetElements`.

This enables simplifications such as
```elm
List.member b ([a,b] ++ rest)
--> True

Set.isEmpty (Set.fromList ([a,b] ++ rest))
--> False
```

## Examples in docs
I am not sure about which/how many cases I should add in the documentation for this change.
I added one for the two cases above, but there are many more that are now included.
Not sure how to generalize it.
Maybe the examples also are a bit complex, and this change doesn't really need any doc examples?

## Multiple fixes
This can lead to multiple fixes, at least when writing tests.
```elm
List.member a ([a,b] ++ [c,d])
--> True
and
--> List.member a ([a,b,c,d])
```

I guess this is unavoidable and also happens for a lot of other rules and that this is fine, but the solutions is probably just write tests that avoid this issue?

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -6839,17 +6839,26 @@ listGetElements resources =
         , \expressionNode ->
             case AstHelpers.removeParens expressionNode of
                 Node _ (Expression.OperatorApplication "++" _ leftList rightList) ->
-                    case ( listGetElements resources leftList, listGetElements resources rightList ) of
-                        ( Just leftElements, Just rightElements ) ->
+                    case
+                        listGetElements resources leftList
+                            |> Maybe.map
+                                (\lefts ->
+                                    ( lefts
+                                    , if lefts.allKnown then
+                                        listGetElements resources rightList
+
+                                      else
+                                        Nothing
+                                    )
+                                )
+                    of
+                        Just ( leftElements, Just rightElements ) ->
                             Just { allKnown = leftElements.allKnown && rightElements.allKnown, known = leftElements.known ++ rightElements.known }
 
-                        ( Just leftElements, Nothing ) ->
+                        Just ( leftElements, Nothing ) ->
                             Just { known = leftElements.known, allKnown = False }
 
-                        ( Nothing, Just rightElements ) ->
-                            Just { known = rightElements.known, allKnown = False }
-
-                        ( Nothing, Nothing ) ->
+                        Nothing ->
                             Nothing
 
                 _ ->
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -6849,7 +6849,7 @@ listGetElements resources =
                         ( Nothing, Just rightElements ) ->
                             Just { known = rightElements.known, allKnown = False }
 
-                        _ ->
+                        ( Nothing, Nothing ) ->
                             Nothing
 
                 _ ->
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -563,6 +563,9 @@ Destructuring using case expressions
     List.member a [ a, b, c ]
     --> True
 
+    List.member a ([a, b] ++ list)
+    --> True
+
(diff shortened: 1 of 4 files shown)


tests/Simplify/ListTest.elm
```diff
@@ -1517,13 +1517,45 @@ a = List.member d (b :: c :: d :: eToZ)
 a = True
 """
                         ]
-        , test "should not report List.member d (a :: b :: c :: dToZ)" <|
+        , test "should not report List.member e (a :: b :: c :: dToZ)" <|
             \() ->
                 """module A exposing (..)
-a = List.member e (b :: c :: d :: eToZ)
+a = List.member e (a :: b :: c :: eToZ)
```
lue-bird, line 1523: We don't use the letter `a` there because `a` already refers to the name of the declaration and we don't want to create a recursive declaration here.

src/Simplify.elm
```diff
@@ -6834,6 +6834,24 @@ listGetElements resources =
...
+                        ( Just leftElements, Just rightElements ) ->
+                            Just { allKnown = leftElements.allKnown && rightElements.allKnown, known = leftElements.known ++ rightElements.known }
+
+                        ( Just leftElements, Nothing ) ->
+                            Just { known = leftElements.known, allKnown = False }
+
+                        ( Nothing, Just rightElements ) ->
+                            Just { known = rightElements.known, allKnown = False }
+
+                        _ ->
```
lue-bird, line 6852: Since we already know this can only be the concrete case `( Nothing, Nothing )` I'd do an exhaustive match

tests/Simplify/SetTests.elm
```diff
@@ -843,6 +843,24 @@ a = Set.isEmpty (Set.singleton x)
                             |> Review.Test.whenFixed """module A exposing (..)
 import Set
 a = False
+"""
+                        ]
+        , test "should replace Set.isEmpty (Set.fromList ([a,b] ++ rest)) by False" <|
+            \() ->
+                """module A exposing (..)
+import Set
+a = Set.isEmpty (Set.singleton x)
```
lue-bird, line 852: This doesn't test what the test name says

tests/Simplify/ListTest.elm
```diff
@@ -1318,7 +1318,7 @@ d = List.member g (List.filter f list)
 """
                     |> Review.Test.run ruleWithDefaults
                     |> Review.Test.expectNoErrors
-        , test "should replace List.member a [] by Nothing" <|
+        , test "should replace List.member a [] by False" <|
```
morteako, line 1321: Small text name fix

tests/Simplify/ArrayTest.elm
```diff
@@ -2046,6 +2046,7 @@ d = Array.get n (Array.repeat m x)
 e = Array.get 0 (Array.repeat m x)
 f = Array.get 0 (Array.initialize m func)
 g = Array.get n (Array.initialize m func)
+h = Array.get 0 (Array.fromList (xs ++ [1]))
```
morteako, line 2049: Added this test to ensure that the listGetElements known elements must be in the correct order

src/Simplify.elm
```diff
@@ -6834,6 +6834,33 @@ listGetElements resources =
...
+                        listGetElements resources leftList
+                            |> Maybe.map
+                                (\lefts ->
+                                    ( lefts
+                                    , if lefts.allKnown then
+                                        listGetElements resources rightList
+
+                                      else
+                                        Nothing
+                                    )
```
morteako, line 6852: To avoid doing redudant work on the right side

lue-bird (approved):

jfmengels (approved):
Thank you @morteako :blush:

PR #289, opened 2024-02-24
Add expectNan to Configuration documentation

morteako:
Seems lik expectNan should also be mentioned, if I understand correctly.

It is mentioned below in the `defaults` documentation


lue-bird (approved):

jfmengels:
Thank you 😊

Issue #157, opened 2023-09-17
Skip conversion step for operations like isEmpty


jfmengels:
```elm
Dict.fromList >> Dict.isEmpty
--> List.isEmpty

Set.fromList >> Set.isEmpty
--> List.isEmpty

✅
Dict.toList >> List.isEmpty
--> Dict.isEmpty

✅
Set.toList >> List.isEmpty
--> Set.isEmpty

✅
Dict.toList >> List.length
--> Dict.size

✅
Set.toList >> List.length
--> Set.size

-- Note: This one doesn't work
Dict.fromList >> Dict.size
--> List.length
```

I'm sure we can find more operations like this for `Dict`/`Set`/`List`. And maybe there are similar things we can do for `Maybe`/`Result`.

morteako:
@jfmengels  Some of these got implemented in : https://github.com/jfmengels/elm-review-simplify/pull/293

The two missing (not including the one that doesn't work 😅 )
```
Dict.fromList >> Dict.isEmpty
--> List.isEmpty

Set.fromList >> Set.isEmpty
--> List.isEmpty
```

jfmengels:
Implemented in #293 and #303 .

PR #303, opened 2024-04-04
 .fromList >> .isEmpty --> List.isEmpty

morteako:
Implements the remaining parts of #157 , and also includes the Array equivalent


jfmengels:
Awesome, thank you @morteako :blush:

Issue #26, opened 2022-07-11
Checking for simplifiable List.length calls (+ Dict/Set)


jfmengels:
:heavy_check_mark: 

**What the Simplify rule should additionally do:**

Check for usages of `List.length` that can be simplified to usages of `List.isEmpty`.
Similarly, we want the same thing for `Dict` and `Set` with the `size` function

**What problems does it solve:**

`List.length` is a lot slower than `List.isEmpty` as it needs to traverse the entire list to get the number of elements, which when compared to a number like 0, is discarded right away.
In some places, using `List.isEmpty` could show significant speed boosts.

I also think that it is easier to understand `List.isEmpty list` than `List.length list >= 1`, as well as being slightly less error-prone (off-by-one, using `>` instead of `>=`, comparing with 0 instead of 1...). Using `List.isEmpty` also makes the code more consistent.

**Example of things the rule would report:**

```elm
List.length list == 0
--> List.isEmpty list ✅

-- All checks should work when inversing the operands
0 == List.length list
--> List.isEmpty list ✅

List.length list /= 0
--> not (List.isEmpty list) ✅

List.length list > 0
--> not (List.isEmpty list)

List.length list >= 1
--> not (List.isEmpty list)

case List.length list of
    0 -> x
    _ -> y
--> if List.isEmpty list then x else y

-- The following are longer in text, but probably more performant (This should actually be benchmarked first)
list == [] ✅
--> List.isEmpty list

list /= [] ✅
--> not (List.isEmpty list)

Array.length array == 0
--> Array.isEmpty array ✅

Dict.size dict == 0
--> Dict.isEmpty dict ✅

Set.size set == 0
--> Set.isEmpty set ✅
```

**Example of things the rule would not report:**

```elm
List.isEmpty list

List.length list -- outside of a comparison

List.length list > 1 -- and other checks that can't easily be translated to `List.isEmpty`

case List.length list of
    0 -> x
    1 -> y
    _ -> z
```

**Requesting feedback:**

Should `List.length list == 2` be replaced by a pattern match like `case list of ; [ _ , _ ] -> True ; _ -> False`?
That would probably be a more performant way to make the check, but I'm not sure the automatic fix would be nice.
EDIT after feedback: No, let's ignore this.

miniBill:
#346 and #344 have implemented some of this

PR #314, opened 2024-07-08
prefer isEmpty functions to checking if length/size == 0 for Set, Dict, and List

perkee:
Partially solve https://github.com/jfmengels/elm-review-simplify/issues/26

Remaining work for this PR
- [x] fix whitespace issue. It sounds like this is non-blocking because elm-format runs afterwards. EDIT: Fixed by @mfonism thank you!!
- ~Maybe reduce the number of tests, idk.~ eh, elm-test is very fast. 

Non blocking work that could be in this PR if we want but it's at best nice-to-have
- [ ] Apply same logic to Strings. The issue is that replacing with `== ""` is a little bit harder.
- [ ] Apply same logic to `List.length >= 1` (and `Set.size` etc).
- [ ] Apply same logic to `List.length < 1` 
- [ ] Throw an error without a fix for `List.length < 0`

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -3874,11 +3874,17 @@ equalityChecks isEqual =
                                     [ Fix.replaceRangeBy
                                         fnRange
                                         replacement
-                                    , Range.combine
-                                        [ Node.range thatNode
-                                        , checkInfo.operatorRange
-                                        ]
-                                        |> Fix.removeRange
+                                    , Fix.removeRange <|
+                                        case Range.compare (Node.range thisNode) (Node.range thatNode) of
+                                            LT ->
+                                                { start = (Node.range thisNode).end
+                                                , end = (Node.range thatNode).end
+                                                }
+
+                                            _ ->
+                                                { start = (Node.range thatNode).start
+                                                , end = (Node.range thisNode).start
+                                                }
                                     ]
                                 )
 
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -3865,12 +3865,7 @@ equalityChecks isEqual =
                         Just { message, details, fnRange, replacement } ->
                             Just
                                 (Rule.errorWithFix { message = message, details = details }
-                                    (Range.combine
-                                        [ checkInfo.leftRange
-                                        , checkInfo.operatorRange
-                                        , checkInfo.rightRange
-                                        ]
-                                    )
+                                    fnRange
                                     [ Fix.replaceRangeBy
                                         fnRange
                                         replacement
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -864,6 +864,12 @@ Destructuring using case expressions
     List.unzip []
     --> ( [], [] )
 
+    List.length l == 0
+    --> List.isEmpty l
+
+    List.length l /= 0
+    --> (not << List.isEmpty) l
+
 
 ### Arrays
 
@@ -1091,6 +1097,12 @@ Destructuring using case expressions
     List.isEmpty (Set.toList set)
     --> Set.isEmpty set
 
+    Set.size set == 0
+    --> Set.isEmpty set
+
+    Set.size set /= 0
+    --> (not << Set.isEmpty) set
+
 
 ### Dict
 
@@ -1183,6 +1195,12 @@ Destructuring using case expressions
     -- The following simplification also works for Dict.keys, Dict.values
     List.isEmpty (Dict.toList dict)
     --> Dict.isEmpty dict
+
+    Set.size dict == 0
+    --> Set.isEmpty dict
+
+    Set.size dict /= 0
+    --> (not << Set.isEmpty) dict
 
 
 ### Cmd / Sub
@@ -3994,7 +4012,7 @@ compareWithZeroChecks checkInfo isEqual node =
                                     newName
 
                                 else
-                                    "(not " ++ newName ++ ")"
+                                    "(not << " ++ newName ++ ")"
                         in
                         Just
                             { message = "This can be replaced with a call to " ++ replacement
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -999,6 +999,9 @@ Destructuring using case expressions
     List.isEmpty (Array.toList array)
     --> Array.isEmpty array
 
+    a = List.length l == 0
+    --> a = List.isEmpty l
+
 
 ### Sets
 
@@ -3834,6 +3837,46 @@ equalityChecks : Bool -> OperatorApplicationCheckInfo -> Maybe (Error {})
 equalityChecks isEqual =
     firstThatConstructsJust
         [ \checkInfo ->
+            let
+                handle : Node Expression -> Node Expression -> Maybe (Error {})
+                handle thisNode thatNode =
+                    case compareWithZeroChecks checkInfo.lookupTable isEqual thisNode of
+                        Just { message, details, fnRange, replacement } ->
+                            Just
+                                (Rule.errorWithFix { message = message, details = details }
+                                    (Range.combine
+                                        [ checkInfo.leftRange
+                                        , checkInfo.operatorRange
+                                        , checkInfo.rightRange
+                                        ]
+                                    )
+                                    [ Fix.replaceRangeBy
+                                        fnRange
+                                        replacement
+                                    , Range.combine
+                                        [ Node.range thatNode
+                                        , checkInfo.operatorRange
+                                        ]
+                                        |> Fix.removeRange
+                                    ]
+                                )
+
+                        Nothing ->
+                            Nothing
+            in
+            case ( Node.value checkInfo.left, Node.value checkInfo.right ) of
+                ( Expression.Integer _, Expression.Integer _ ) ->
+                    Nothing
+
+                ( Expression.Integer 0, _ ) ->
+                    handle checkInfo.right checkInfo.left
+
+                ( _, Expression.Integer 0 ) ->
+                    handle checkInfo.left checkInfo.right
+
+                _ ->
+                    Nothing
+        , \checkInfo ->
             findMap
                 (\side ->
                     if Evaluate.getBoolean checkInfo side.node == Determined isEqual then
@@ -3912,6 +3955,64 @@ equalityChecks isEqual =
                 Normalize.Unconfirmed ->
                     Nothing
         ]
+
+
+compareWithZeroChecks :
+    ModuleNameLookupTable
+    -> Bool
+    -> Node Expression
+    ->
+        Maybe
+            { message : String
+            , details : List String
+            , fnRange : Range
+            , replacement : String
+            }
+compareWithZeroChecks lookupTable isEqual node =
+    let
+        maybeRangeAndNames =
+            [ ( "List.isEmpty", "List.length", Fn.List.length )
+            , ( "Dict.isEmpty", "Dict.size", Fn.Dict.size )
+            , ( "Set.isEmpty", "Set.size", Fn.Set.size )
+
+            -- , ( "String.isEmpty", "String.length", Fn.String.length ) is this the best replacement? Should it be == ""?
+            ]
+                |> List.map (\( newName, oldName, fn ) -> ( newName, oldName, AstHelpers.getSpecificFnCall fn lookupTable node ))
+                |> List.filterMap
+                    (\( newName, oldName, maybeCall ) ->
+                        Maybe.map
+                            (\call ->
+                                { range = call.fnRange
+                                , newName = newName
+                                , oldName = oldName
+                                }
+                            )
+                            maybeCall
+                    )
+                |> List.head
+    in
+    case maybeRangeAndNames of
+        Just { range, oldName, newName } ->
+            let
+                replacement =
+                    if isEqual then
+                        newName
+
+                    else
+                        "(not " ++ newName ++ ")"
+            in
+            Just
+                { message = "This can be replaced with a call to " ++ replacement
+                , details =
+                    [ oldName ++ " takes as long to run as the list is long"
+                    , "whereas " ++ newName ++ " runs in constant time."
+                    ]
+                , fnRange = range
+                , replacement = replacement
+                }
+
+        Nothing ->
+            Nothing
 
 
 
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -4017,6 +4017,7 @@ compareWithZeroChecks checkInfo isEqual node =
                             newName =
                                 qualifiedToString (qualify newFn checkInfo)
 
+                            replacement : String
                             replacement =
                                 if isEqual then
                                     newName
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1005,8 +1005,11 @@ Destructuring using case expressions
     List.isEmpty (Array.toList array)
     --> Array.isEmpty array
 
-    a = List.length l == 0
-    --> a = List.isEmpty l
+    Array.length a == 0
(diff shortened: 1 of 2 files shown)


tests/Simplify/EqualTest.elm
```diff
@@ -934,6 +934,587 @@ a = ({ a = 1 }).a == ({ a = 1 }).b
...
+                                    , "whereas Set.isEmpty runs in constant time."
+                                    ]
+                                , under = "size s == 0"
+                                }
+                                |> Review.Test.whenFixed """module A exposing (..)
+import Set as CoreSet exposing (..)
+
+
+a : Bool
+a = isEmpty s
```
perkee, line 1168: How do I get the best name for the package as it is imported here? Replacing with `CoreSet.isEmpty` is also fine (preferable, even) but `Set.isEmpty` would be wrong here.

As a real world example, consider a codebase where is used `rtfeldman/elm-sorter-experiment`. Then in a file you would have

```elm
import Sort.Set as Set -- Because you are creating a lot of sets of custom types
import Set as CoreSet -- for the one time you make a set of integers.
```

tests/Simplify/EqualTest.elm
```diff
@@ -934,6 +934,587 @@ a = ({ a = 1 }).a == ({ a = 1 }).b
...
+                            { message = "This can be replaced with a call to Set.isEmpty"
+                            , details =
+                                [ "Set.size takes as long to run as the list is long"
+                                , "whereas Set.isEmpty runs in constant time."
+                                ]
+                            , under = "0 == Set.size s"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a : Bool
+a =  Set.isEmpty s
```
perkee, line 1189: I was torn about spacing here. When I do my text replacement I could try to move one to the left in the range, but that assumes that the code is `elm-format`ted. Is there a canonical way to replace this? Am

tests/Simplify/EqualTest.elm
```diff
@@ -934,6 +934,587 @@ a = ({ a = 1 }).a == ({ a = 1 }).b
...
+                            { message = "This can be replaced with a call to List.isEmpty"
+                            , details =
+                                [ "List.length takes as long to run as the list is long"
+                                , "whereas List.isEmpty runs in constant time."
+                                ]
+                            , under = "(l |> List.length) == 0"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a : Bool
+a = (l |> List.isEmpty)
```
perkee, line 1008: what's the best way to replace this here? I think this should eliminate the pizza and instead replace with `List.isEmpty l` but maybe that's a job for another rule?

src/Simplify.elm
```diff
@@ -3914,6 +3957,64 @@ equalityChecks isEqual =
...
+            , replacement : String
+            }
+compareWithZeroChecks lookupTable isEqual node =
+    let
+        maybeRangeAndNames =
+            [ ( "List.isEmpty", "List.length", Fn.List.length )
+            , ( "Dict.isEmpty", "Dict.size", Fn.Dict.size )
+            , ( "Set.isEmpty", "Set.size", Fn.Set.size )
+
+            -- , ( "String.isEmpty", "String.length", Fn.String.length ) is this the best replacement? Should it be == ""?
```
perkee, line 3978: Yeah what's the best replacement for Strings?

jfmengels:
Hi @perkee, thanks for taking a look at this! :heart: 

> Should we include String and if so what is the best replacement == "" or String.isEmpty

Yeah good question. I think maybe replacing by `== ""` is better here. It looks the most like the code before (using an equality sign).

> How do I handle white space?

We rely on `elm-format` applying afterwards, so don't let it bother you too much. In general, we try to make the minimum amount of changes when fixing to get the desired result. That helps with keeping the formatting similar (avoid introducing new lines, or collapsing onto one line if possible) and to avoid removing comments.

There's sometimes a trade-off with making a nice diff for the user, so that they understand what the fix is doing. Overall, don't let it bother you too much.

tests/Simplify/EqualTest.elm
jfmengels: If I'm not mistaken, you'd do that like this:
```elm
qualifiedToString (qualify ( [ "Set" ], "isEmpty" ) checkInfo)
```

src/Simplify.elm
```diff
@@ -999,6 +999,9 @@ Destructuring using case expressions
     List.isEmpty (Array.toList array)
     --> Array.isEmpty array
 
+    a = List.length l == 0
+    --> a = List.isEmpty l
```
jfmengels, line 1003: For consistency:
```suggestion
    List.length list == 0
    --> List.isEmpty list
```

Although this fits better in the `## Lists` section, so it should probably be moved

tests/Simplify/EqualTest.elm
jfmengels: This is honestly fine. If you really want to improve it, you probably should:
- if `List.length l == 0`: remove from the end of `List.length` until the end of `0`
- if `0 == List.length l`: remove from the start of `0` until the start of `List.length`

and then also replace `List.length` by the appropriate function. Let me know if I misunderstood your question.

tests/Simplify/EqualTest.elm
jfmengels: Yeah that should be another rule. I think this is as it should be :ok_hand:

src/Simplify.elm
```diff
@@ -3914,6 +3957,64 @@ equalityChecks isEqual =
...
+                            newName : String
+                            newName =
+                                qualifiedToString (qualify newFn checkInfo)
+
+                            replacement =
+                                if isEqual then
+                                    newName
+
+                                else
+                                    "(not " ++ newName ++ ")"
```
perkee, line 3997: this is wrong! `not : Bool -> Bool` but we want something like `invert : (a -> Bool) -> Bool -> Bool` which doesn't exist. So I'll need to do some work rearranging the parentheses.

perkee: this was very easy to fix lol. `<<` ezpz

perkee:
@mfonism hero

perkee:
@jfmengels remaining items are all better candidates for future PRs, so this is ready for the big time.

tests/Simplify/EqualTest.elm
```diff
@@ -934,6 +934,787 @@ a = ({ a = 1 }).a == ({ a = 1 }).b
...
+                            { message = "This can be replaced with a call to (not << List.isEmpty)"
+                            , details =
+                                [ "Whereas List.length takes as long to run as the number of elements in the List,"
+                                , "List.isEmpty runs in constant time."
+                                ]
+                            , under = "0 /= (l |> List.length)"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a : Bool
+a = (l |> (not << List.isEmpty))
```
jfmengels, line 1114: I have to admit I don't like this `not <<` fix. I think it would be much better to wrap in a `not` call, and ideally in a way that follows the existing format:
- `List.length l /= 0` -> `not (List.isEmpty l)`
- `(l |> List.length) /= 0` -> `(l |> List.isEmpty |> not)`
- `(List.length <| l) /= 0` -> `(not <| List.length <| l)`

As far as I can tell, the original position of the operands (`... /= 0` or `0 /= ...`) does not impact the fix.

As for the error message, I think the following would fit:

> This can be replaced with a call to `List.empty` and `not`

The "and `not`" part would only be there when encountering `/=`.

What do you think?

tests/Simplify/EqualTest.elm
```diff
@@ -934,6 +934,787 @@ a = ({ a = 1 }).a == ({ a = 1 }).b
...
+a : Bool
+a = 0 /= (l |> List.length)
+"""
+                    |> Review.Test.run ruleWithDefaults
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "This can be replaced with a call to (not << List.isEmpty)"
+                            , details =
+                                [ "Whereas List.length takes as long to run as the number of elements in the List,"
+                                , "List.isEmpty runs in constant time."
```
jfmengels, line 1108: Every element in the details list is a paragraph. Here, I think it's better for the two sentences to be in one paragraph (so a list with only a single string containing the whole message).

tests/Simplify/EqualTest.elm
```diff
@@ -934,6 +934,787 @@ a = ({ a = 1 }).a == ({ a = 1 }).b
...
+
+
+a : Bool
+a = isEmpty s
+"""
+                        ]
+        , test "should replace 0 == Set.size s with Set.isEmpty s" <|
+            \() ->
+                """module A exposing (..)
+a : Bool
```
jfmengels, line 1174: There's technically a missing import here (and the same in other tests).

tests/Simplify/EqualTest.elm
```diff
@@ -934,6 +934,787 @@ a = ({ a = 1 }).a == ({ a = 1 }).b
...
+"""
+                    |> Review.Test.run ruleWithDefaults
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "This can be replaced with a call to (not << Set.isEmpty)"
+                            , details =
+                                [ "Whereas Set.size takes as long to run as the number of elements in the Set,"
+                                , "Set.isEmpty runs in constant time."
+                                ]
+                            , under = "Set.size s /= 0"
```
jfmengels, line 1248: I think the range should be put on the smallest section of the code that makes sense. Here, `s` could potentially be a large expression, on multiple lines.

I don't know exactly what would be best.
It could be only under `/= 0`. Though if this is on multiple lines, then it should only be only under `/=`.
But we could also put it under `Set.size`, and - unless someone has a stronger opinion on this - I think it should be there.

tests/Simplify/EqualTest.elm
perkee: good call. Will update.

tests/Simplify/EqualTest.elm
perkee: roger! Was wondering how that all worked. Will replace with one line.

tests/Simplify/EqualTest.elm
```diff
@@ -934,6 +934,787 @@ a = ({ a = 1 }).a == ({ a = 1 }).b
...
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a : Bool
+a = (d |> (not << Dict.isEmpty))
+"""
+                        ]
+        ]
+
+
+arrayTests : Test
```
jfmengels, line 1522: Rename to `arrayIsEmptyTests`?

perkee: good call

tests/Simplify/EqualTest.elm
```diff
@@ -934,6 +934,787 @@ a = ({ a = 1 }).a == ({ a = 1 }).b
...
+arrayTests =
+    describe "Array.length should be Set.isEmpty"
```
jfmengels, line 1524:
```suggestion
    describe "Array.length should be Array.isEmpty"
```

(Same issue for the dict tests at least)

tests/Simplify/EqualTest.elm
```diff
@@ -934,6 +934,787 @@ a = ({ a = 1 }).a == ({ a = 1 }).b
...
+                                , "isEmpty runs in constant time."
+                                ]
+                            , under = "size s == 0"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+import Set as CoreSet exposing (..)
+
+
+a : Bool
+a = isEmpty s
```
jfmengels, line 1168: I think you wanted to have a test where you'd end up having to insert `CoreSet.isEmpty`, but you ended up not having one.

Maybe add a test with this code:
```elm
import Set as CoreSet exposing (size)

a : Bool
a = size s == 0
```

perkee: Thought of this over breakfast this morning actually.

perkee: I spend a lot of time testing your helper functions rather than the meat of my replacement, so that's part of the whole test pruning thing I think I want to do.

tests/Simplify/EqualTest.elm
perkee: This pushes me more to rework the tests generally. These are very repetitive (obviously). I should instead rotate through the pairs of replaced/replacement functions inside of one list of tests. I should also vary the collection expressions and not just go replacing single letter/word expressions but rather something larger.

tests/Simplify/EqualTest.elm
perkee: is there an existing rule and replacement out there that handles this kind of replacement I could look at?

tests/Simplify/EqualTest.elm
jfmengels: I think it's fine the way you did it, errors like this are bound to happen, but once they're fixed (a nice IDE helps with that), the tests are really nice, even if they're repetitive. At least in my opinion, you're allowed to disagree :sweat_smile:

tests/Simplify/EqualTest.elm
jfmengels: Feel free to add comments to the helper functions if you think they're unclear!
I'm not sure whether you need to do pruning, it's okay to have multiple similar tests :man_shrugging:

tests/Simplify/EqualTest.elm
jfmengels: I haven't found one so far. @lue-bird does this ring a bell to you?

jfmengels:
> Throw an error without a fix for List.length list < 0

This should be replaced by `True` always (or `False` for `>= 0`). I think we already have similar simplifications for `List.isEmpty` and similar functions.
(Again, not necessary for this PR)

tests/Simplify/EqualTest.elm
perkee: Sorry, I meant that the bulk of the runtime of these tests spends time on nuances of the helper functions. No, they were incredibly helpful and clear!!

tests/Simplify/EqualTest.elm
lue-bird: mapOnWrappedChecks and wrapperMapNChecks have something like that
```elm
case checkInfo.callStyle of
      Pipe LeftToRight ->
          [ Fix.insertAt elementCountCall.parentRange.end
              (" |> " ++ qualifiedToString (qualify Fn.Basics.not checkInfo))
          ]

      Pipe RightToLeft ->
          [ Fix.insertAt elementCountCall.parentRange.start
              (qualifiedToString (qualify Fn.Basics.not checkInfo) ++ " <| ")
          ]

      Application ->
          [ Fix.insertAt elementCountCall.parentRange.end ")"
          , Fix.insertAt elementCountCall.parentRange.start (qualifiedToString (qualify Fn.Basics.not checkInfo) ++ " (")
          ]
```

perkee: This is really solid, thanks. It's going to take me a little bit of time since I don't have the `callStyle` of the function application here. I may have to add it to `AstHelpers.getSpecificFnCall` in order to have it in this location. The parent `checkInfo` is about the operator application of `==` so that doesn't have it, naturally.

jfmengels:
@perkee My sincere apologies for not tending this PR earlier.

I have made it so that the fixes don't introduce `not <<`, which removes the only thing I had issues with this PR.

Thank you so much @perkee and @mfonism for working on this, this is going to be a great addition to the rule!
(and I'll try to be more reactive if you decide to help out again)

Issue #115, opened 2023-08-12
simplifications with multiple `<`, `<=`, `>=`, `>`


lue-bird:
**Simplifications we can add:**
```elm
a <= b && a >= b
--> a == b

a < b || a > b
--> a /= b

a < b && a > b
--> False

a < b && a >= b
--> False

a <= b && a > b
--> False

a <= b || a >= b
--> True

a <= b || a > b
--> True

a < b || a >= b
--> True
```

All of them also work for `a` and or `b = NaN`, see [this test runner ellie](https://ellie-app.com/nCXZVZLqx4Ja1).

I'm sure there are more cases.

**What problems does it solve:**
This will catch sneaky bugs like
```elm
rowIsInRange : Range -> Int -> Bool
rowIsInRange range =
    \row ->
        (row >= range.start.row)
            && (row <= range.start.row)
```
or at least simplify more complex conditions.

jfmengels:
I think these sound good! The `not`should probably be pretty easy to do too :+1: 

> Do you agree that a < b || a > b --> a /= b is always a clear simplification to the user?

I think so. Reading `a /= b` is at least much easier to read (and I guess to understand what it really means) than `a < b || a > b`.

But I think it might impact performance negatively to change. Let's take the similar example of `a <= b && a >= b` to `a == b` and look at the compiled JS code.
Before:
```js
-- a <= b && a >= b compiles down to
a <= b && a >= b
-- which is really fast
```
After:
```js
-- a == b compiles down to
_Utils_eq(a, b)
-- which is a bit slower. Ideally we want to compile this to
a === b
```

If one of the operands is a number `a <= 0 && a >= 0` then we will get `a === 0` which will be fast. So we should probably benchmark this before we simplify this expressions to using `==` or `/=` where it wasn't already. I hope that in the next version of Elm the compilation will retain type information and be able to optimize this better.

In https://github.com/jfmengels/elm-review-simplify/pull/31 we made it so that we can know about past conditions (in the same expression or in the scope). I wonder if we could re-use that and whether that will influence the proposed simplifications (related changelog: https://github.com/jfmengels/elm-review-simplify/blob/main/CHANGELOG.md#2017---2022-08-14).

lue-bird:
I tried `elm make`, `elm make --optimize` and `elm-optimize-level-2` but I always get a call to `_Utils_cmp` for comparison operators

```js
function _Utils_cmp(x, y, ord) {
    if (typeof x !== "object") {
[...]

PR #322, opened 2024-09-29
Add configuration option doNotSimplifyMicroOptimizedCode

lue-bird:
The [introduction article for simplify](https://jfmengels.net/simplify/#automatic-fixes-for-unused-constructors) presented snowballing simplifications that includes the step `prefix ++ "" → prefix` – which simplify doesn't do till this day due to performance considerations.

I've argued that the vast majority of users do not and should not care about that.
For those folks it's much more helpful to offer a fix for `++ ""` since that's very likely not intentional.
A similar case is https://github.com/jfmengels/elm-review-simplify/issues/115#issuecomment-1676434266

This PR adds the option `Simplify.doNotSimplifyMicroOptimizedCode` which very similar to `expectNaN` is by default set to provide all simplifications it can.

### DIFF
--- a/tests/Simplify/PlusPlusTest.elm
+++ b/tests/Simplify/PlusPlusTest.elm
@@ -1,6 +1,7 @@
 module Simplify.PlusPlusTest exposing (all)
 
 import Review.Test
+import Simplify
 import Test exposing (Test, describe, test)
 import TestHelpers exposing (ruleWithDefaults)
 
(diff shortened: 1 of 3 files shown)


tests/Simplify/PlusPlusTest.elm
```diff
@@ -45,13 +46,34 @@ a = "a" ++ ""
 a = "a"
 """
                         ]
-        , test """should not report x ++ "" (because this can lead to better performance)""" <|
+        , test """should not report x ++ "" when doNotSimplifyMicroOptimizedCode is enabled (because this can lead to better performance)""" <|
```
jfmengels, line 49: In practice, we could fine-tune this kind of situation, even with micro-optimizations allowed.

```elm
x ++ "" -- report, it's useless
x ++ y ++ "" -- Don't report, it's useful
"foo" ++ y ++ "" -- report, it's useless
```

jfmengels:
I'm on the fence on this one. I think the idea is good, that in many cases people won't care. When the rule reports an issue around something that was not meant to be optimized (that ended up with `++ ""` due to other simplifications), this should be a positive thing.

But I don't know when I would recommend using this new option. For libraries, recommending this could make the library slower.

For applications, I don't think it's fine-grained enough. In applications, we mostly don't care about performance,because most code is not in a hot path. But there will be sections of the application where performance matters very much.

So I don't think that the function should be a boolean. Maybe it should a list of modules - or maybe of function names - where `Simplify` will not apply the simplifications removing the micro-optimizations.

What do you think?

---

Also, I think I would be in favor of adding in the error details how to disable the rule from reporting the issue (by enabling the option), as I imagine that not knowing about the new option could lead to some frustration for maintainers that do want their code to use these tricks. Linking to the function is probably good enough. That said, since the rule is usually ran with `--fix-all`, I don't know if people will even see the link...

Issue #326, opened 2024-12-13
Remove duplicate entries in Set.fromList/Dict.fromList


jfmengels:
If some of the keys are the same, then we can remove some of them (potentially telling the user there's a problem with their setup).

```elm
Dict.fromList [ ( 1, a ), ( 2, b ), ( 3, c ), ( 1, d ) ]
-->
Dict.fromList [ ( 2, b ), ( 3, c ), ( 1, d ) ]

Dict.fromList [ ( 1, a ), ( 2, b ), ( 3, c ) ]
  |> Dict.insert 1 d
-->
Dict.fromList [ ( 2, b ), ( 3, c ) ]
  |> Dict.insert 1 d
--> Or alternatively put it directly in the list, probably nicer
Dict.fromList [ ( 2, b ), ( 3, c ), ( 1, d ) ]


Set.fromList [ 1, 2, 3, 1 ]
-->
Set.fromList [ 2, 3, 1 ]

Set.fromList [ 1, 2, 3 ]
  |> Set.insert 1
-->
Set.fromList [ 1, 2, 3 ]
```

lue-bird:
We can also implement these checks for unknown key expressions like
```elm
Set.fromList [ a, a ]
--> Set.fromList [ a ]
```

NaNs do not follow this rule
```elm
Set.fromList [ 0/0, 0/0 ]
--> Set.fromList [NaN,NaN]
```
so we should only do it if `expectNaN` is configured False.


--------

I think I'll implement some of those

lue-bird:
Another simplification we can do is
```elm
Dict.fromList [ a, anything, a ]
--> Dict.fromList [ anything, a ]
```
(again, doesn't work when expecting NaN)

PR #328, opened 2024-12-22
Dict.fromList & Set.fromList duplicate key merge

lue-bird:
implements #326 except for `insert`

```elm
Set.fromList [ a, a ] --> Set.fromList [ a ]
Dict.fromList [ ( a, v0 ), ( a, v1 ) ] --> Dict.fromList [ ( a, v1 ) ]
```

Honestly I'm surprised we didn't implement this already because it seems pretty useful.

The check for duplicates is `n^2` but I guess that shouldn't be too bad because these explicit `fromList` calls are rare and usually not terribly long in non-codegen code.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -6168,13 +6168,29 @@ dictFromListChecks =
                     case Node.value checkInfo.firstArg of
                         Expression.ListExpr elements ->
                             let
+                                knownEntryTuples : List { entryRange : Range, first : Node Expression }
+                                knownEntryTuples =
+                                    elements
+                                        |> List.filterMap
+                                            (\entry ->
+                                                case AstHelpers.getTuple2 checkInfo.lookupTable entry of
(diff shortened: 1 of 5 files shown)


src/Simplify.elm
```diff
@@ -5998,6 +6004,69 @@ setFromListChecks =
         [ intoFnCheckOnlyCall (emptiableFromListChecks setCollection)
         , wrapperFromListSingletonChecks setCollection
         , onSpecificFnCallReturnsItsLastArgCheck Fn.Set.toList
+        , intoFnCheckOnlyCall
+            (\checkInfo ->
+                if checkInfo.expectNaN then
```
jfmengels, line 6009: Later improvement: I guess we could do this check but only for literals if `expectNaN` is `True`.

jfmengels:
This looks great, thanks @lue-bird!

PR #329, opened 2024-12-28
Improvements to fromList key deduplication simplification

jfmengels:
I noticed there was a bug where some entries would be removed: We did `Dict.fromList [ ( key, a ), thing, ( key, b ) ] -> Dict.fromList [ ( key, b ) ]` which is incorrectly removing `thing`.

Other improvements:
- The `allDifferent` checks now aborts early when finding an error
- Normalization is done only once instead of once for each comparison with a prior key and once for when it's compared to later keys.

PR can be reviewed commit by commit.

### DIFF
--- a/tests/Simplify/DictTest.elm
+++ b/tests/Simplify/DictTest.elm
@@ -367,6 +367,26 @@ import Dict
 a = Dict.fromList [ ( key, v1 ) ]
 """
                         ]
+        , test "should replace Dict.fromList [ ( key, v0 ), thing, ( key, v1 ) ] by Dict.fromList [ thing, ( key, v1 ) ]" <|
+            \() ->
+                """module A exposing (..)
+import Dict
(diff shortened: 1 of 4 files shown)


tests/Simplify/DictTest.elm
```diff
@@ -365,6 +365,26 @@ a = Dict.fromList [ ( key, v0 ), ( key, v1 ) ]
                             |> Review.Test.whenFixed """module A exposing (..)
 import Dict
 a = Dict.fromList [ ( key, v1 ) ]
+"""
+                        ]
+        , test "should replace Dict.fromList [ ( key, v0 ), thing, ( key, v1 ) ] by Dict.fromList [ thing, ( key, v1 ) ]" <|
```
lue-bird, line 370: maybe also add the same test for Set?
Otherwise really good refactors and fixes.
I keep missing cases like that so I'll try to test on more examples next time :)

lue-bird (approved):

PR #330, opened 2025-01-06
More duplicate checks for Set/Dict.fromList

jfmengels:
More duplicate checks from #326.

- `Dict.fromList [ x, x ]` now gets simplified to `Dict.fromList [ x ]`
- The duplicate checks now work even if expecting NaN when the Dict keys or Set values are literals, so `Set.fromList [ 0, 0 ]` now gets simplified.


src/Simplify/AstHelpers.elm
```diff
@@ -950,6 +952,74 @@ isSpecificUnappliedBinaryOperation symbol checkInfo expression =
...
+                        (operator == "+")
+                            || (operator == "-")
+                            || (operator == "*")
+                            || (operator == "/")
+                            || (operator == "^")
+                    then
+                        True
+
+                    else
+                        canEqualOrContainNaNHelp rest
```
lue-bird, line 993: for one, the left and right expressions should be added to the recursive call. For 2, operators `|>` and `<|` can produce NaN

src/Simplify/AstHelpers.elm
```diff
@@ -1013,7 +1013,37 @@ canEqualOrContainNaNHelp nodes =
...
+                    canEqualOrContainNaNHelp rest
+
+                Expression.Hex _ ->
+                    canEqualOrContainNaNHelp rest
+
+                Expression.Floatable _ ->
+                    canEqualOrContainNaNHelp rest
+
+                Expression.Negation _ ->
+                    canEqualOrContainNaNHelp rest
```
lue-bird, line 1033: this buddy can be NaN if the negated expression can be NaN.
Making it exhaustive was udeful after all :)

src/Simplify/AstHelpers.elm
```diff
@@ -950,6 +952,102 @@ isSpecificUnappliedBinaryOperation symbol checkInfo expression =
...
+                        canEqualOrContainNaNHelp (left :: right :: rest)
+
+                Expression.FunctionOrValue _ _ ->
+                    True
+
+                Expression.RecordAccess _ _ ->
+                    True
+
+                Expression.RecordUpdateExpression _ _ ->
+                    True
```
lue-bird, line 1000: can never be NaN

src/Simplify/AstHelpers.elm
```diff
@@ -950,6 +952,102 @@ isSpecificUnappliedBinaryOperation symbol checkInfo expression =
...
+
+                Expression.CaseExpression _ ->
+                    True
+
+                Expression.LambdaExpression _ ->
+                    True
+
+                Expression.RecordExpr _ ->
+                    True
```
lue-bird, line 1012: can never be NaN

src/Simplify/AstHelpers.elm
```diff
@@ -950,6 +952,102 @@ isSpecificUnappliedBinaryOperation symbol checkInfo expression =
...
+
+                Expression.LetExpression _ ->
+                    True
+
+                Expression.CaseExpression _ ->
+                    True
+
+                Expression.LambdaExpression _ ->
+                    True
```
lue-bird, line 1009: can never be NaN

src/Simplify/AstHelpers.elm
```diff
@@ -950,6 +952,102 @@ isSpecificUnappliedBinaryOperation symbol checkInfo expression =
...
+
+                Expression.RecordAccess _ _ ->
+                    True
+
+                Expression.RecordUpdateExpression _ _ ->
+                    True
+
+                Expression.LetExpression _ ->
+                    True
```
lue-bird, line 1003: instead check if the result can be NaN

src/Simplify/AstHelpers.elm
```diff
@@ -950,6 +952,102 @@ isSpecificUnappliedBinaryOperation symbol checkInfo expression =
...
+
+                Expression.CaseExpression _ ->
+                    True
```
lue-bird, line 1006: instead check if any case result can be NaN

jfmengels:
Good catches :+1:

lue-bird (commented):
- record update still needs to be skipped
- Any reason for the rename? I was hoping to reuse this to make checks like multiplying by 0 work more often even when expecting NaN.

jfmengels:
> record update still needs to be skipped

:+1: 

> Any reason for the rename? 

The name felt clunky, that it didn't properly match the usage. I don't have a problem with it being renamed now, or later when it's used for more purposes. I still am not 100% satisfied by the current name anyway.

lue-bird (approved):

lue-bird:
I just realized a mistake in what I suggested:
While record and record update can't directly be NaN, they can obviously return it!
I'll fix that in my current PR

jfmengels:
> While record and record update can't directly be NaN, they can obviously contain it!

That is true, but records are not comparable, so it doesn't matter (unless we start looking at the individual fields of the record). That's part of why I added `This will return `False` for expressions that are known to not be `comparable` (e.g. records) nor `numbers`` comment, and also why I renamed it from the "canEqualOrContain" name, because containing it is not relevant for records wrt keys.

PR #331, opened 2025-01-09
Allow some expressions even when expecting nan

lue-bird:
Thanks to the new "could be NaN" helper introduced in #330 , we can enable some existing checks even when expecting NaN, for example
```elm
-- expectNaN enabled
List.member 0 [ 0, 1 ] --> True
```

Bonus:
- fixes the helper as commented in https://github.com/jfmengels/elm-review-simplify/pull/330#issuecomment-2579724094
- gives the helper a more general name
- upgrades elm-review bc it didn't work

### DIFF
--- a/src/Simplify/AstHelpers.elm
+++ b/src/Simplify/AstHelpers.elm
@@ -13,7 +13,7 @@ module Simplify.AstHelpers exposing
     , moduleNameFromString, qualifiedName, qualifiedModuleName, qualifiedToString
     , declarationListBindings, letDeclarationListBindings, patternBindings, patternListBindings
     , nameOfExpose
-    , isPotentialNaNKey
+    , couldBeValueContainingNaN
     )
 
 {-|
@@ -59,7 +59,7 @@ module Simplify.AstHelpers exposing
 
 @docs declarationListBindings, letDeclarationListBindings, patternBindings, patternListBindings
 @docs nameOfExpose
-@docs isPotentialNaNKey
+@docs couldBeValueContainingNaN
 
 -}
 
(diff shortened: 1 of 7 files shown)


src/Simplify/AstHelpers.elm
```diff
@@ -952,32 +952,36 @@ isSpecificUnappliedBinaryOperation symbol checkInfo expression =
...
 
 
-{-| Indicates whether this value is potentially NaN in the context of a Dict key or Set value, meaning that it could return `False` when `==` with itself.
+{-| Indicates whether this value is potentially NaN,
+meaning that it could return `False` when `==` with itself.
 
-This will return `False` for expressions that are known to not be `comparable` (e.g. records) nor `numbers`.
+This will return `False` for expressions that are known to
+only contain literals (e.g. [ ( 0, { name = "string" } ) ]),
+are known operations that never produce NaN (e.g. `a ++ b`),
```
jfmengels, line 960: `a ++ b` can produce NaN because you could have `[ NaN ] ++ []` which is itself comparable.

jfmengels:
I'll re-post the comment from the previous merged PR (https://github.com/jfmengels/elm-review-simplify/pull/330#issuecomment-2579826680), because it's harder to have a conversation in a closed PR 😅 

> While record and record update can't directly be NaN, they can obviously contain it!

That is true, but records are not comparable, so it doesn't matter (unless we start looking at the individual fields of the record). That's part of why I added `This will return `False` for expressions that are known to not be `comparable` (e.g. records) nor `numbers`` comment, and also why I renamed it from the "canEqualOrContain" name, because containing it is not relevant for records wrt comparable keys.

---

For the purposes of comparable keys, but also for multiplying by 0 (where values have to be comparable to type check), records can be completely ignored. For `List.member` however, they are indeed mandatory. I'm wondering whether we should have a boolean argument to the function (or simply have a different function) indicating whether we're in a context where values are comparable, in which case we can safely ignore records.

lue-bird:
> For the purposes of comparable keys, but also for multiplying by 0 (where values have to be comparable to type check), records can be completely ignored. For List.member however, they are indeed mandatory. I'm wondering whether we should have a boolean argument to the function (or simply have a different function) indicating whether we're in a context where values are comparable, in which case we can safely ignore records.

I don't see a reason to as comparable values and numbers will never be records in the first place

PR #333, opened 2025-01-09
copy List.member checks for Set.member and Dict.member

### LINKED ISSUE #332: member on known collection not containing the given element
Currently we simplify
```elm
List.member 0 [ 2, 0, 1 ] --> True
```
but not the other way around:
```elm
List.member -999 [ 2, 0, 1 ] --> False
```

This works for `List`, `Set`, `Dict` and has the usual NaN restrictions.

I'll try to implement this unless someone else wants to.

lue-bird:
Fully implements/closes #332:
```elm
Set.member x (Set.fromList [ y, x ]) --> True
Set.member -999 (Set.fromList [ 0, 1 ]) --> False
Dict.member x [ ( y, v0 ), ( x, v1 ) ] --> True
Dict.member -999 [ ( 0, v0 ), ( 1, v1 ) ] --> False
List.member -999 [ 0, 1 ] --> False
```
Bonus:
- copy the singleton checks from List.member for Set.member
  ```elm
  Set.member x (Set.singleton y) --> x == y
  ```

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -587,6 +587,9 @@ Destructuring using case expressions
     List.member a [ b ]
     --> a == b
 
+    List.member -999 [ 0, 1 ]
+    --> False
+
     List.map f [] -- same for most List functions like List.filter, List.filterMap, ...
     --> []
 
@@ -853,7 +856,6 @@ Destructuring using case expressions
     List.sort [ a ]
     --> [ a ]
 
-
     -- same for up to List.map5 when any list is empty
     List.map2 f xs []
     --> []
@@ -1029,6 +1031,15 @@ Destructuring using case expressions
     Set.member x Set.empty
     --> False
 
+    Set.member x (Set.singleton y)
+    --> x == y
+
+    Set.member x (Set.fromList [ y, x ])
+    --> True
+
+    Set.member -999 (Set.fromList [ 0, 1 ])
+    --> False
+
     Set.toList Set.empty
     --> []
 
@@ -1116,6 +1127,12 @@ Destructuring using case expressions
     --> 0
 
     Dict.member x Dict.empty
+    --> False
+
+    Dict.member x [ ( y, v0 ), ( x, v1 ) ]
+    --> True
+
+    Dict.member -999 [ ( 0, v0 ), ( 1, v1 ) ]
     --> False
 
     Dict.remove k Dict.empty
@@ -6109,7 +6126,15 @@ setSizeChecks =
 
 setMemberChecks : IntoFnCheck
 setMemberChecks =
-    intoFnCheckOnlyCall (collectionMemberChecks setCollection)
+    intoFnCheckOnlyCall
+        (firstThatConstructsJust
+            [ callOnEmptyReturnsCheck
+                { resultAsString = \res -> qualifiedToString (qualify Fn.Basics.falseVariant res) }
+                setCollection
+            , knownMemberChecks setCollection
+            , wrapperMemberChecks setCollection
+            ]
+        )
 
 
 setInsertChecks : IntoFnCheck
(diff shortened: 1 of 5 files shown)


src/Simplify.elm
```diff
@@ -6290,7 +6315,22 @@ dictSizeChecks =
...
+            [ callOnEmptyReturnsCheck
+                { resultAsString = \res -> qualifiedToString (qualify Fn.Basics.falseVariant res) }
+                dictCollection
+            , knownMemberChecks
+                { represents = "dict"
+                , elements =
+                    { countDescription = "size"
+                    , elementDescription = "key"
+                    , determineCount = dictDetermineSize
+                    , get = dictGetKeys
```
jfmengels, line 6329: This representation of Dict is defined twice, maybe we should have two distinct Dict collection top-level variables?

jfmengels: Nevermind, I thought I saw this twice.

tests/Simplify/DictTest.elm
```diff
@@ -771,6 +771,78 @@ a = Dict.remove k Dict.empty
                             |> Review.Test.whenFixed """module A exposing (..)
 import Dict
 a = Dict.empty
+"""
+                        ]
+        , test "should replace Dict.member 1 (Dict.fromList [ ( 0, () ), ( 1, () ), ( 2, () ) ]) by True" <|
+            \() ->
+                """module A exposing (..)
+import Dict
+a = Dict.member 1 (Dict.fromList [ ( 0, () ), ( 1, () ), ( 2, () ) ])
```
jfmengels, line 780: Oh wait, this means we now have the capabilities to simplify the following 😍 

```elm
Dict.get 1 (Dict.fromList [ ( 0, () ), ( 1, () ), ( 2, () ) ])
-> Just ( 1, () )

Dict.get 1 (Dict.fromList [ ( 0, () ), ( 2, () ) ])
-> Nothing
```

tests/Simplify/DictTest.elm
```diff
@@ -771,6 +771,78 @@ a = Dict.remove k Dict.empty
                             |> Review.Test.whenFixed """module A exposing (..)
 import Dict
 a = Dict.empty
+"""
```
jfmengels, line 774: These new tests are under `Dict.remove`, which is the wrong location.

tests/Simplify/DictTest.elm
```diff
@@ -771,6 +771,78 @@ a = Dict.remove k Dict.empty
...
+                    |> Review.Test.run ruleExpectingNaN
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Dict.member on a dict which does not contain the given key will result in False"
+                            , details = [ "You can replace this call by False." ]
+                            , under = "Dict.member"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+import Dict
+a = False
```
jfmengels, line 845: Another simplification we could do, is to transform `Dict.member x (Dict.fromList [ ( key, value ), reference ])` to `List.member x [ key, Tuple.first reference ]`. I'm not 100% convinced yet, but it could remove references.

tests/Simplify/DictTest.elm
```diff
@@ -771,6 +771,78 @@ a = Dict.remove k Dict.empty
...
+"""
+                        ]
+        , test "should replace Dict.member 0 (Dict.fromList [ ( 2, () ), ( 3, () ), ( 1, () ) ]) by False when expecting NaN" <|
```
jfmengels, line 830: Could you these tests (unless I missed them):
- Simplification to `False` does not occur when there is a non-literal key and it is otherwise not found? Example: `Dict.member 0 (Dict.fromList [ ( 2, () ), ( 3, () ), ( reference, () ) ])` should not get simplified. We could however potentially simplify to `Dict.member 0 (Dict.fromList [ ( reference, () ) ])` as some entries can be ignored.
- If the needle is potentially NaN, then we should report depending on NaN. So `Dict.member reference (Dict.fromList [ ( 2, () ), ( 3, () ), ( reference, () ) ])` is simplified to `True` when expectNaN is False, and ignored when expectNaN is True. I think this is handled, but it would be nice to have tests for it.

Issue #342, opened 2025-05-23
Dict.union (Dict.singleton, Dict.singleton)


gampleman:
**What the rule should do:**

```elm

Dict.union (Dict.singleton "foo" "bar") (Dict.singleton "some" other)

-->

Dict.fromList [( "foo", "bar" ), ( "some", other )]

---

Dict.union (Dict.singleton "foo" "bar") (Dict.fromList [("some", other)])

-->


Dict.fromList [( "foo", "bar" ), ( "some", other )]

```

this already happens for `Dict.fromList`, but this expands it to work with Dict.singleton. Basically anywhere we use `Dict.fromList`, we should also work with `Dict.singleton`.

gampleman:
We could also potentially do something with the fact that `Dict.union (Dict.singleton key value) dict)` is equivalent to `Dict.insert key value dict`.

lue-bird:
Right. so additionally:
```elm
Dict.union (Dict.singleton k v) dict
--> Dict.insert k v dict

Set.union (Set.singleton a) set
--> Set.insert a set
Set.union set (Set.singleton a)
--> Set.insert a set
Set.union (Set.singleton a) (Set.singleton b)
--> Set.fromList [ a, b ]
```

PR #343, opened 2025-05-24
Add Dict and Set union with singleton simplifications

lue-bird:
The uncontroversial half of the simplifications suggested in https://github.com/jfmengels/elm-review-simplify/issues/342:
```elm
Set.union (Set.singleton a) set
--> Set.insert a set
Set.union << Set.singleton
--> Set.insert

Set.union set (Set.singleton a)
--> Set.insert a set

Dict.union (Dict.singleton k v) dict
--> Dict.insert k v dict
Dict.union << Dict.singleton k
--> Dict.insert k
```

`Set.union set << Set.singleton` is not simplified because I don't know of a way to fix it without creating a lambda or using flip.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+The rule now simplifies:
+- `Set.union (Set.singleton a) set` to `Set.insert a set`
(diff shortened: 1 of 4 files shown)


jfmengels:
Great work once again @lue-bird, thank you! :relaxed:

Issue #344, opened 2025-06-03
Comparison with empty list should use List.isEmpty


jfmengels:
```elm
x == []
--> List.isEmpty x
[] == x
--> List.isEmpty x
x /= []
--> not (List.isEmpty x)
```

jfmengels:
Closed by https://github.com/jfmengels/elm-review-simplify/pull/346

PR #346, opened 2025-06-06
Fixes Comparison with empty list should use List.isEmpty #344

miniBill:


lue-bird (approved):
Looks good. Some extras would be

- simplifying parenthesized `[]`
- simplifying `(==) []` → `List.isEmpty`. Needs to be inserted here: https://github.com/jfmengels/elm-review-simplify/blob/main/src/Simplify.elm#L2615
- making it generic for e.g. `Array.empty`, `Set.empty`, `Dict.empty`, `""`

We could also potentially mention pattern matching as an alternative in the error details for beginner elm users.

but someone (or me) can also do all/some of that in a separate PR

jfmengels:
Good suggestions!
I'll merge this in the meantime then. @miniBill would you like to have a look at implementing those suggestions?
And thanks again 😊

PR #348, opened 2025-06-07
Add checks for comparisons with empty for Array, Dict, Set

miniBill:
Implements some of the suggestions from #346

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -3985,83 +3985,7 @@ equalityChecks isEqual =
 
                 Normalize.Unconfirmed ->
                     Nothing
-        , \checkInfo ->
-            let
-                surroundWith : Node Expression -> ( String, String )
-                surroundWith (Node _ expr) =
-                    let
-                        fnName : String
-                        fnName =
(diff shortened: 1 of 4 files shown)


tests/Simplify/StringTest.elm
```diff
@@ -71,6 +71,118 @@ a = String.isEmpty "a"
...
+        , test "should replace comparisons to the empty string with String.isEmpty (reverse order)" <|
+            \() ->
+                """module A exposing (..)
+
+a = "" == x"""
+                    |> Review.Test.run ruleWithDefaults
+                    |> Review.Test.expectErrors
+                        [ Review.Test.error
+                            { message = "Comparison with an empty string can be replaced by a call to String.isEmpty"
+                            , details = [ "You can replace this comparison to an empty string with a call to String.isEmpty, which is more efficient." ]
```
jfmengels, line 83: I'm honestly not sure that it is more efficient for strings.

I do wonder if we'll get some complaints for this string simplification, I myself am not sure I like it that much :thinking:

jfmengels:
I dropped the changes for `""`.
Thank you @miniBill :relaxed:

PR #350, opened 2025-06-08
Compare with empty

### LINKED ISSUE #349: Simplify `(==) []` to `List.isEmpty`
From @lue-bird's comment on a [previous issue](https://github.com/jfmengels/elm-review-simplify/pull/346#pullrequestreview-2907129592):
> ` (==) [] → List.isEmpty`. Needs to be inserted here: https://github.com/jfmengels/elm-review-simplify/blob/main/src/Simplify.elm#L2615

miniBill:
Fixes #349

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -4147,27 +4147,6 @@ comparisonWithEmptyChecks isEqual checkInfo =
             else
                 ( "not (" ++ fnName ++ " ", ")" )
 
-        isEmpty : Node Expression -> Maybe ModuleName
-        isEmpty node =
-            case node of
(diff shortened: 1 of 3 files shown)


src/Simplify.elm
```diff
@@ -4220,6 +4221,30 @@ comparisonWithEmptyChecks isEqual checkInfo =
                     Nothing
 
 
+{-| If the expression is [], String.empty
```
jfmengels, line 4224: String.empty? Did you mean Array.empty and friends?

miniBill: Yes. Yes I did

PR #353, opened 2025-06-12
Avoid record updates in critical path

miniBill:


miniBill:
This provides a performance improvement of about 15% on a large codebase

jfmengels:
Thank you 😊

PR #356, opened 2025-06-12
Use an ExpressionDict to speed up duplicate checks in Set.fromList

miniBill:


jfmengels:
I'd be happy to merge this, but I would first like to know whether this is faster than the simpler alternative of encoding with `elm-syntax` (or a faster version that doesn't care about nodes) that was discussed in https://discord.com/channels/534524278847045633/729776405524512789/1383214581945729157.

PR #357, opened 2025-06-12
Simplify List.folds with Dict/Set/List insert to fromList

caseyWebb:
Replaces instances of

```elm
List.foldl Set.insert Set.empty items
    -> Set.fromList items

List.foldl (\(k, v) -> Dict.insert) Dict.empty items
    -> Dict.fromList items
```


caseyWebb:
Yanked the conditionals, fixed the Dict case, added support for lambdas 🎉

PR #360, opened 2025-06-20
Speed up duplicate checks by hashing values

jfmengels:
Supersedes https://github.com/jfmengels/elm-review-simplify/pull/356 as it does the same thing but ends up being faster.

I did a rather naive hashing of the AST. There may be some ambiguity but I *think* it's good enough. I'll (later) try this on a work repo before merging it as a sanity check.
cc @miniBill

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1485,6 +1485,7 @@ import Set exposing (Set)
 import Simplify.AstHelpers as AstHelpers exposing (emptyStringAsString, qualifiedToString)
 import Simplify.CallStyle as CallStyle exposing (FunctionCallStyle)
 import Simplify.Evaluate as Evaluate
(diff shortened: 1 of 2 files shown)


miniBill:
I think it _might_ be faster to use `List String` rather than `String` as a key

jfmengels:
@miniBill Maybe, but the memory consumption would be higher I believe and the equality check would be slower.

In any case, I can't on my not-so-great benchmarking setup (running it against my work project) not notice a very measurable difference between this version and the one on `main` (or against the versions you had). If you ever get the chance, I'd love it if you could try it out.

jfmengels:
I'm merging this. We can always undo or improve it further later. Thank you for the help @miniBill :relaxed:

Issue #359, opened 2025-06-20
String.fromInt 2 => "2"


miniBill:
Same for `String.fromFloat <float>` and `String.toInt "string"`.

_Maybe_ also for `String.toFloat "float"` but I'm wary of IEEE here.

jfmengels:
I agree that `String.toFloat` may be slightly dangerous, though we could compare it to the original string and see if it's indeed the same. The same it true for integers, which may be really large.

So to summarize in our proposal style, and add a few more:
```elm
String.fromInt 2
--> "2"

String.fromFloat 2.2
--> "2.2"

String.toInt "23"
--> Just 23

String.toInt "23345678987654345678987654345678"
-- no change, does not result in the same output as in the literal

String.toInt "string"
--> Nothing

String.fromChar 'c'
--> "c"
```

Not sure about this one. What do you think?
```elm
String.cons 's' "tr"
--> "str"
```

jfmengels:
Adding it here from #362:
I'm not sure whether we want this or not:
```elm
String.fromInt 0x2
--> "2"
```

PR #362, opened 2025-07-11
String.fromInt/fromFloat simplifications

miniBill:
Draft because I'd like to cover all of #359

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -449,6 +449,9 @@ Destructuring using case expressions
     String.foldl (\_ soFar -> soFar) initial string
     --> initial
 
+    String.fromInt 123
+    --> "123"
+
 
 ### Maybes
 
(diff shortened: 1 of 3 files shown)


lue-bird (approved):

src/Simplify.elm
```diff
@@ -5113,6 +5121,54 @@ stringFromListChecks =
         ]
 
 
+stringFromIntChecks : IntoFnCheck
+stringFromIntChecks =
+    intoFnCheckOnlyCall
+        (\checkInfo ->
+            case Node.value checkInfo.firstArg of
+                Expression.Integer i ->
```
jfmengels, line 5129: Should we do this as for `Hex` values?

jfmengels:
Thanks for working on this (and sorry for the delay, was on vacation) :relaxed:

PR #364, opened 2025-10-16
Add simplification for String.join singleton

### LINKED ISSUE #336: String.join / String.concat with a single element
**What the rule should do:**
`String.join "foo" [ a ] => a`
`String.concat [ a ] => a`

**Example of things the rule would not report:**
```elm
String.join "foo" [ "bar", "baz" ]
String.concat [ "bar", "baz" ]
```

Part of Simplify

**I am looking for:**
- Someone to implement it with/for me

miniBill:
Fixes #336

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -373,6 +373,9 @@ Destructuring using case expressions
     String.join "" list
     --> String.concat list
 
+    String.join "," [ var ]
+    --> var
+
     String.length "abc"
     --> 3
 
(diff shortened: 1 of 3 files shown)


src/Simplify.elm
```diff
@@ -11978,6 +11982,36 @@ callOnEmptyReturnsCheck config collection checkInfo =
...
+                            checkInfo.fnRange
+                            [ Fix.removeRange
+                                { start = checkInfo.parentRange.start
+                                , end = singleRange.start
+                                }
+                            , Fix.removeRange
+                                { start = singleRange.end
+                                , end = checkInfo.parentRange.end
+                                }
+                            ]
```
lue-bird: This will e.g. fix
```elm
[ f <| string ] |> String.join sep
--> f <| string
```
which can lead to invalid syntax.
A better alternative is `replaceBySubExpressionFix`, see https://github.com/jfmengels/elm-review-simplify/blob/main/src/Simplify.elm#L14642-L14648

Adding a test for this would be nice, too

Issue #352, opened 2025-06-10
Identify more things as the identity function


jfmengels:
We have a lot of simplifications that depend on a function being `identity`.
We currently treat the `Basics.identity` function and functions like `\x -> x` as the same.

But there are more patterns like that. I just discovered this one in a project:
```elm
List.map (\( p, e ) -> ( p, e )) list
--> list
```

---

There's also
```elm
(\(Wrapper w) -> Wrapper w)
```
at least when `Wrapper` does not have a phantom type (otherwise this might lead to a compilation error).

Although simplifying this one to `identity` (or like other identity simplifications) might remove some compiler reminders when you add new a new variant next to `Wrapper`. I'm therefore not sure about this one yet.

---

Also, if any top-level is equivalent to `identity`, should we then consider applying `identity` simplifications?

```elm
newIdentity x = x

List.map newIdentity list
--> list ?
```

---

Could there be others?

PR #369, opened 2025-12-08
recognize more patterns and expressions as identical

lue-bird:
Implements https://github.com/jfmengels/elm-review-simplify/issues/352. In effect, this enables simplifying for example
```elm
Maybe.map (\(a, b) -> (a, b))
```
Includes:
```elm
\() -> ()
\a -> a
\(a) -> ((a))
\(a,b) -> (a,b)
\(a,b,c) -> (a,b,c)
\(a as unused) -> a
\(unused as a) -> a
\({a,b} as r) -> {r|b=b,a=a}
```
and compositions of these.
Does not include `\(Variant a b) -> Variant a b` due to the issues with phantom types etc discussed in the issue referenced above.

### DIFF
--- a/src/Simplify/AstHelpers.elm
+++ b/src/Simplify/AstHelpers.elm
@@ -236,7 +236,7 @@ getSpecificFnCall :
             , callStyle : FunctionCallStyle
             }
 getSpecificFnCall ( moduleName, name ) lookupTable expressionNode =
-    case getValueOrFnOrFnCall expressionNode of
+    case getValueOrFnOrFnCall lookupTable expressionNode of
         Just call ->
             case call.args of
                 firstArg :: argsAfterFirst ->
@@ -265,7 +265,8 @@ getSpecificFnCall ( moduleName, name ) lookupTable expressionNode =
 {-| Parse a value or the collapsed function or a lambda fully reduced to a function
 -}
 getValueOrFnOrFnCall :
-    Node Expression
+    ModuleNameLookupTable
+    -> Node Expression
     ->
         Maybe
             { nodeRange : Range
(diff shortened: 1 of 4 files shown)


src/Simplify/AstHelpers.elm
```diff
@@ -647,16 +653,208 @@ getReducedLambda expressionNode =
...
+                                        _ ->
+                                            False
+
+                                _ ->
+                                    False
+                    in
+                    matchesRecordUpdate
+                        || expressionReconstructsIrrefutablePattern lookupTable
+                            unparenthesizedExpressionNotLocalReference
+                            unparenthesizedAliasedPattern
```
jfmengels: TCO gets disabled with `||`. I suggest this gets rewritten as the following:
```suggestion
                    if matchesRecordUpdate then
                        True

                    else
                        expressionReconstructsIrrefutablePattern lookupTable
                            unparenthesizedExpressionNotLocalReference
                            unparenthesizedAliasedPattern
```

src/Simplify/AstHelpers.elm
```diff
@@ -647,16 +653,208 @@ getReducedLambda expressionNode =
...
+
+        Pattern.TuplePattern [ patternPart0, patternPart1 ] ->
+            case getTuple2 lookupTable expressionNode of
+                Nothing ->
+                    False
+
+                Just expressionParts ->
+                    expressionReconstructsIrrefutablePattern lookupTable expressionParts.first patternPart0
+                        && expressionReconstructsIrrefutablePattern lookupTable expressionParts.second patternPart1
```
jfmengels: Similarly, this would enable a bit more TCO. I don't know how much we want to rewrite code to make this more TCO, as it does make things a bit less readable.
```suggestion
                    if expressionReconstructsIrrefutablePattern lookupTable expressionParts.first patternPart0 then
                        True

                    else
                        expressionReconstructsIrrefutablePattern lookupTable expressionParts.second patternPart1
```

src/Simplify/AstHelpers.elm
```diff
@@ -647,16 +653,208 @@ getReducedLambda expressionNode =
...
+            --     unsafeMap : (Int -> Int) -> IntIsNat isNat -> IntIsNat isNat
+            --
+            --     abs : IntIsNat Never -> IntIsNat ()
+            --     abs integer = integer |> unsafeMap Basics.abs |> unsafeMarkNat
+            --
+            -- Notice how `unsafeMarkNat` has different input and output
+            -- type and is _not equivalent to identity_.
+            -- This specific example is arguably contrived but it does
+            -- occur in real life code!
+            -- So there is nothing we can do here, surprisingly (phantom types suck)
```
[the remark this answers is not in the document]
jfmengels: I just read my own comment from the issue

> Although simplifying this one to identity (or like other identity simplifications) might remove some compiler reminders when you add new a new variant next to Wrapper. I'm therefore not sure about this one yet.

I'm still not sure about it, but right now I would lean towards saying simplifying this is okay?

PR #373, opened 2025-12-13
Improve performance

lue-bird:
Various performance improvements. The more performant code should not be noticeably less readable but if you think certain commits fail in that regard, please point them out.
Running it with `elm-benchmark` using `Review.Rule.reviewV3`, the code after the PR was consistently around 35% faster than before but I'd guess the actual overall improvement is closer to 10-30%.

If you have other ideas to try, I'm happy to implement them.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1715,7 +1715,7 @@ type alias ModuleContext =
     , moduleBindings : Set String
     , localBindings : RangeDict (Set String)
     , branchLocalBindings : RangeDict (Set String)
-    , rangesToIgnore : RangeDict ()
+    , rangesToIgnore : List Range
     , rightSidesOfPlusPlus : RangeDict ()
     , customTypesToReportInCases : Set ( ModuleName, ConstructorName )
     , localIgnoredCustomTypes : List Constructor
@@ -1873,7 +1873,7 @@ fromProjectToModule =
             , moduleBindings = Set.empty
             , localBindings = RangeDict.empty
             , branchLocalBindings = RangeDict.empty
-            , rangesToIgnore = RangeDict.empty
+            , rangesToIgnore = []
             , rightSidesOfPlusPlus = RangeDict.empty
             , localIgnoredCustomTypes = []
             , customTypesToReportInCases = projectContext.customTypesToReportInCases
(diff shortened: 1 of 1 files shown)


src/Simplify.elm
```diff
@@ -6622,29 +6618,27 @@ listSortWithChecks =
...
+                                    operationWithFirstArgIsEquivalentToFnError
                                         { firstArgDescription = "a comparison that always returns LT"
                                         , replacementFn = Fn.List.reverse
                                         }
                                         checkInfo
-                                    )
 
-                            EQ ->
-                                Just fixToIdentity
+                                EQ ->
```
jfmengels: Should we use `_ ->` here to not have duplicate branches?

src/Simplify/CoreHelpers.elm
```diff
@@ -422,8 +422,8 @@ listMapToStringsThenJoinAfter soFar elementToString separator list =
 {-| Like `Maybe.andThen` but for the `Nothing` case.
 Exceptionally useful for trying multiple things in order
 -}
-maybeOnNothing : (() -> Maybe a) -> Maybe a -> Maybe a
-maybeOnNothing nextTry maybe =
+onNothing : (() -> Maybe a) -> Maybe a -> Maybe a
+onNothing nextTry maybe =
```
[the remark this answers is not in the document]
lue-bird: orElse seems too generic as an exposed helper name and I also _really_ like the naming scheme used by e.g. [Task.onError](https://dark.elm.dmy.fr/packages/elm/core/latest/Task#onError)

src/Simplify.elm
```diff
@@ -15619,30 +15619,27 @@ alwaysResultsInConstantError :
...
-            if config.replacementNeedsParens then
-                "(" ++ string ++ ")"
-
-            else
-                string
-
         replacement : QualifyResources {} -> String
-        replacement =
+        replacement res =
             case checkInfo.argCount - (1 + List.length checkInfo.argsAfterFirst) of
```
jfmengels: Probably worth extracting this value we're pattern matching on, so we don't recompute it twice (since `replacement` is called twice).

lue-bird: It's called with different resources, not sure there's much we can do

src/Simplify.elm
```diff
@@ -2353,133 +2353,178 @@ expressionVisitor node config context =
...
+
+                Just error ->
+                    [ error ]
+            , { rangesToIgnore =
+                    case expressionChecked.subRangeToIgnore of
+                        Nothing ->
+                            context.rangesToIgnore
+
+                        Just additionalRangeToIgnore ->
+                            additionalRangeToIgnore :: context.rangesToIgnore
```
jfmengels: Good catch!

PR #374, opened 2025-12-15
Support simplifying ++ "" in more places

jfmengels:
Seems like we were supporting `"a" ++ ""` -> `"a"`, but if we were to add anything before or after `x ++ "a" ++ "" ++ y` then the simplification would not work.

It was fairly easy to change the code to report an empty string anywhere, but trickier to add the exception to keep `a ++ ""` to enable optimizations (although now that I'm writing this, we should only do that if there is a concatenation, like `a ++ b ++ ""`).

The way I'm doing it is by starting from a string literal  then go to the right. If I find at least one other string literal, then it means I can safely remove one without defeating the `++ ""` optimization.


jfmengels:
Yeah, it took me a while to figure this out as well :sweat_smile:

PR #375, opened 2025-12-16
More operation on operation with equal arguments simplifications

lue-bird:
```elm
String.left n (String.left n str)
--> String.left n str

String.right n (String.right n str)
--> String.right n str

List.filter f (List.filter f list)
--> List.filter f list

List.take n (List.take n list)
--> List.take n list

Array.filter f (Array.filter f array)
--> Array.filter f array

Array.set i v1 (Array.set i v0 array)
--> Array.set i v1 array

Set.insert k (Set.insert k set)
--> Set.insert k set

Set.remove k (Set.remove k set)
--> Set.remove k set

Set.filter f (Set.filter f set)
--> Set.filter f set

Dict.insert k v1 (Dict.insert k v0 dict)
--> Dict.insert k v1 dict

Dict.remove k (Dict.remove k dict)
--> Dict.remove k dict

Dict.filter f (Dict.filter f dict)
--> Dict.filter f dict
```
including the same checks for composition.
Similar simplifications exist for `union` but they are different enough to be in a separate PR

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -7367,7 +7367,7 @@ dictMemberChecks =
 
 dictInsertChecks : IntoFnCheck
 dictInsertChecks =
-    operationMakesPreviousOperationWithEqualFirstArgUnnecessaryCheck { firstArgDescription = "key" }
+    operationOverridesPreviousOperationWithEqualFirstArgCheck { firstArgDescription = "key" }
 
 
 dictRemoveChecks : IntoFnCheck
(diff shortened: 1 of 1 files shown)


src/Simplify.elm
```diff
@@ -11533,6 +11546,96 @@ setChecks collection =
...
+
+    array |> Array.set 1 0 |> Array.set 1 100
+    --> array |> Array.set 1 100
+
+Use `operationDoesNotChangeResultOfOperationCheck`
+when _all arguments_ need to be the same, even if there is just one argument
+
+-}
+operationMakesPreviousOperationWithEqualFirstArgUnnecessaryCheck : { firstArgDescription : String } -> IntoFnCheck
+operationMakesPreviousOperationWithEqualFirstArgUnnecessaryCheck config =
```
jfmengels: Alternative name suggestion that I find slightly easier to read, but choose what you prefer: `operationOverridesPreviousOperationWithEqualFirstArg`

jfmengels (approved):
These simplifications feel so obvious... that I'm surprised I never thought of them or seen them suggested before. Great catch!

PR #376, opened 2025-12-16
Concatenate string literals on the same line

jfmengels:
WHile running the preview rule on some code, I discovered code like `String.fromInt 16 ++ "px"` get converted to `"16" ++ "px"` but never get simplified to `"16px"`, which I found surprising.

So this implements the concatenation of these strings. But only if the end of the left side is on the same line as the start of the right side, which prevents getting very large lines like this:

```elm
"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
  ++ "bbbbbbbbbbbbbbbbbbbbbbbb"
  ++ "ccccccccccccccccccccccccccccccc"
--> ❌ we don't want this
"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaabbbbbbbbbbbbbbbbbbbbbbbbccccccccccccccccccccccccccccccc"
```

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -313,7 +313,7 @@ Destructuring using case expressions
     Tuple.second ( a, b )
     --> b
 
-    Tuple.second (Tuple.mapFirst changeSecond tuple)
+    Tuple.second (Tuple.mapFirst changeFirst tuple)
     --> Tuple.second tuple
 
     Tuple.second (Tuple.mapBoth changeFirst changeSecond tuple)


tests/Simplify/PlusPlusTest.elm
```diff
@@ -134,6 +134,176 @@ a = "" ++ "a"
...
+                            , under = "++"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = "ab"
+"""
+                        ]
+        , test """should replace "a" ++ "b" ++ x by "ab" ++ x""" <|
+            \() ->
+                """module A exposing (..)
+a = "a" ++ "b"
```
lue-bird: missing the ++ x

src/Simplify.elm
```diff
@@ -4356,6 +4321,107 @@ findNextStringLiteral (Node range node) previousNode =
...
+                    (Rule.errorWithFix
+                        { message = checkInfo.operator ++ " on " ++ stringCollection.represents ++ " literals can be turned into a single " ++ stringCollection.represents ++ " literal"
+                        , details = [ "Try moving all the elements into a single " ++ stringCollection.represents ++ " literal." ]
+                        }
+                        checkInfo.operatorRange
+                        ([ Fix.removeRange
+                            { start = { row = leftRange.end.row, column = leftRange.end.column - sizeOfQuotes leftUsesTripleQuotes }
+                            , end = { row = rightRange.start.row, column = rightRange.start.column + sizeOfQuotes rightUsesTripleQuotes }
+                            }
+                         ]
```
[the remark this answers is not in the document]
jfmengels: `"""abc""""` is a syntax error (unterminated quote)
`"""abc"""""` is a type error, as it's considered like `"""abc""" ""`, so I don't think we have to handle this.
`""""de"""` is valid (the fourth `"` is a plain quote. So `"a" ++ """"d"""` will result in `"""a"d"""` which is valid.

tests/Simplify/PlusPlusTest.elm
jfmengels: And I've now introduced a new bug. Adding a test case and fixing it.

PR #377, opened 2025-12-17
Simplify calls with comparable values

lue-bird:
A more complete, useful and correct way of parsing comparable expressions, and using them to simplify e.g.
```elm
[ 1.2 ] < [ 1.0, 2.0 ] --> GT
```
Full list of additional simplifications:
```elm
Basics.min n n --> n
Basics.min 3 4 --> 3
Basics.max n n --> n
Basics.max 3 4 --> 4
Basics.compare n n --> EQ -- when `expectNaN` is not enabled
Basics.compare 3 4 --> LT
n < n --> False
n > n --> False -- when `expectNaN` is not enabled
n <= n --> False -- when `expectNaN` is not enabled
n >= n --> False
Basics.truncate 23.4 --> 23 (same for `round`, `floor`, `ceiling`)
```

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -3675,10 +3675,10 @@ intoFnChecks =
     , ( Fn.Basics.not, ( 1, basicsNotChecks ) )
     , ( Fn.Basics.negate, ( 1, basicsNegateChecks ) )
     , ( Fn.Basics.toFloat, ( 1, basicsToFloatChecks ) )
-    , ( Fn.Basics.round, ( 1, intToIntChecks Basics.round ) )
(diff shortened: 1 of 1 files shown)


src/Simplify.elm
```diff
@@ -5373,10 +5373,11 @@ basicsToFloatChecks =
         )
 
 
-intToIntChecks : IntoFnCheck
-intToIntChecks =
+intToIntChecks : (Float -> Int) -> IntoFnCheck
+intToIntChecks operation =
```
jfmengels: I guess this should be renamed, maybe something like `toIntChecks` or `conversionToIntChecks`?

PR #378, opened 2025-12-19
expand "f a a --> a" checks to any equal inner call values across the two arguments and composition

lue-bird:
For `Basics.max`, `Basics.min`, `Set.union`, `Dict.union`, `Set.intersect`, `Dict.intersect` (should apply to any "extremum" function that has the same input and output types), we previously simplified:
```elm
f a a --> a
```
This rule can actually be expanded to cover e.g.
```elm
f (f a d) (f (f a c) b) --> f d (f (f a c) b)
f a >> f (f (f a c) b) --> f (f (f a c) b)
```
or any arbitrarily nested call or composition. Kind of tricky!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,8 +26,10 @@ The rule now simplifies:
 - `String.left n (String.left n string)` to `String.left n string`
 - `String.right n (String.right n string)` to `String.right n string`
 - `Basics.min n n` to `n`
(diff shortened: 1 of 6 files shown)


jfmengels:
Really nice! Thank you :heart: 

> This rule can actually be expanded to cover e.g.

Just so I understand correctly, it currently doesn't cover this. Correct?

lue-bird:
>Just so I understand correctly, it currently doesn't cover this. Correct?

This PR implemented them

jfmengels:
I figured the first example was included, but wasn't sure about the second. Very nice work!

PR #386, opened 2025-12-23
Optimize creation of fixes

jfmengels:
The order of fixes doesn't matter, so we might as well write what's fastest in terms of List creation.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -4169,8 +4169,8 @@ minusChecks checkInfo =
                 , details = [ "You can replace this operation by the negated right number you subtracted from 0, like `-n`." ]
                 }
                 checkInfo.operatorRange
-                (replaceBySubExpressionFix checkInfo.parentRange checkInfo.right
-                    ++ [ Fix.insertAt checkInfo.parentRange.start "-" ]
+                (Fix.insertAt checkInfo.parentRange.start "-"
+                    :: replaceBySubExpressionFix checkInfo.parentRange checkInfo.right
                 )
             )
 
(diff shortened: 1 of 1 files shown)


lue-bird (approved):
Makes sense. I was always afraid to change these because:
- order does matter in rare cases (e.g. insert at a location, then remove from that location to another can be different from removing that range, then inserting at the start, try e.g. in dictMemberOnFromListToListAnyChecks on composition, moving earlier removeRange last)
- sometimes, a specific order can be more intuitive than another, at a small cost as fixes are rarely a performance bottleneck

These don't seem to matter with these changes, so :+1:

jfmengels:
> order does matter in rare cases (e.g. insert at a location, then remove from that location to another can be different from removing that range, then inserting at the start, try e.g. in dictMemberOnFromListToListAnyChecks on composition, moving earlier removeRange last)

Fixes are applied in a way that the order shouldn't matter. If you try to insert at a position that will get removed by `Fix.removeRange`, then that should be considered as a fix collision and the fix won't be applied (and tests should fail).

When I started moving these around I did get different results, which I would consider a bug in `elm-review`, but could not reproduce afterwards. If you find that the order does matter in practice, then let me know.

> sometimes, a specific order can be more intuitive than another, at a small cost as fixes are rarely a performance bottleneck

Absolutely. I've moved these around very mechanically. Feel free to swap things around if they make it easier to read. But I agree, fixes are not going to be a bottleneck.

jfmengels:
Huh. I tried bumping it indeed, and then I couldn't reproducing even when un-bumping the dependencies... weird...

Yes, let's bump it to latest, as well as elm-syntax. That should help users get updated with all dependencies.

I can bump it a bit later today if you wish

PR #390, opened 2025-12-23
Disable merging of lists when there might be structure

jfmengels:
Disabled `List.concat` simplifications that merged `List.concat` without structure.
```elm
grid =
    List.concat
        [ [ O, X, X ]
        , [ O, O, X ]
        , [ X, O, O ]
        ]
--> previously
grid =
    [ O, X, X, O, O, X, X, O, O ]
```
This simplification made it sometimes harder to understand the data, and is therefore getting removed.

It will however still apply when there doesn't seem to be any structure, i.e. when the list is one line or when all sub-items are on different lines.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -14687,30 +14687,39 @@ collectionUnionWithLiteralsChecks config operationInfo collection checkInfo =
         Just literalListSecond ->
             case fromListGetLiteral collection checkInfo.lookupTable operationInfo.first of
                 Just literalListFirst ->
-                    let
-                        fromListLiteralDescription : String
-                        fromListLiteralDescription =
-                            constructionFromListOnLiteralDescription collection.fromList
-                    in
-                    Just
-                        (Rule.errorWithFix
-                            { message = operationInfo.operation ++ " on " ++ fromListLiteralDescription ++ "s can be turned into a single " ++ fromListLiteralDescription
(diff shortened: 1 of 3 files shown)


jfmengels:
That is a very weird error during the publish test...
I'll ignore it for now, since the rest of the tests pass, but I hope we won't encounter it anymore.

src/Simplify.elm
```diff
@@ -6556,25 +6556,34 @@ listConcatChecks =
             (\checkInfo ->
                 case fromListGetLiteral listCollection checkInfo.lookupTable checkInfo.firstArg of
                     Just listLiteral ->
-                        if List.all AstHelpers.isListLiteral listLiteral.elements then
+                        let
+                            (Node firstArgRange _) =
+                                checkInfo.firstArg
+                        in
+                        if
+                            (firstArgRange.start.row /= firstArgRange.end.row)
```
lue-bird: I'd prefer
```elm
not (List.all (\(Node itemRange _) -> itemRange.end.row == firstArgRange.start.row) listLiteral.elements)
```
instead of
```elm
(firstArgRange.start.row /= firstArgRange.end.row)
```
to account for e.g.
```elm
List.concat
    [ [ 1, 2 ]
    ]
```
A test for that case would be nice as well

lue-bird (approved):

Issue #308, opened 2024-04-30
Simplify Dict.keys/values >> List.fold


jfmengels:
Similar to changes like the one introduced in https://github.com/jfmengels/elm-review-simplify/pull/76, but for Dict.

The reason why we have this for `Set` and `Array` but not for `Dict`, is because the fold functions for `Dict` take an additional argument, which requires changes potentially far away. This is however easily resolvable when the fold function is an anonymous function:

When using `Dict.keys`:
```elm
dict
        |> Dict.keys
        |> List.foldl (\key acc -> ...) x
-->
dict
        |> Dict.foldl (\key _ acc -> ...) x
```

When using `Dict.values`:
```elm
dict
        |> Dict.values
        |> List.foldl (\value acc -> ...) x
-->
dict
        |> Dict.foldl (\_ value acc -> ...) x
```

(if `List.foldr` is used, then we would use `Dict.foldr` instead)

PR #391, opened 2025-12-25
Add List.fold on Dict.keys/values simplifications

lue-bird:
Implements/closes https://github.com/jfmengels/elm-review-simplify/issues/308
```elm
-- same for foldr
List.foldl (\v s -> f v s) init (Dict.values dict)
--> Dict.foldl (\_ v s -> f v s) init dict

-- same for foldr
List.foldl (\k s -> f k s) init (Dict.keys dict)
--> Dict.foldl (\k _ s -> f k s) init dict
```
including composition.

I'm very slightly unsure about the rare fix:
```elm
List.foldl f init (Dict.keys dict)
--> Dict.foldl (always << f) init dict
```
I think it could be a bit unintuitive to folks.

Wish you a happy holiday :)

### DIFF
--- a/tests/Simplify/ListTest.elm
+++ b/tests/Simplify/ListTest.elm
@@ -4252,6 +4252,23 @@ a = List.foldl (\\a s -> f a s) init (Dict.values dict)
 a = Dict.foldl (\\_ a s -> f a s) init dict
 """
                         ]
+        , test "should replace List.foldl (\\a -> f a) init (Dict.values dict) by Dict.foldl (\\_ a -> f a) init dict" <|
+            \() ->
+                """module A exposing (..)
+a = List.foldl (\\a -> f a) init (Dict.values dict)
+"""
+                    |> Review.Test.run ruleWithDefaults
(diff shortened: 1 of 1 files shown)


tests/Simplify/ListTest.elm
```diff
@@ -4233,6 +4233,159 @@ a = Array.toList >> List.foldl f x
...
+                            , under = "List.foldl"
+                            }
+                            |> Review.Test.whenFixed """module A exposing (..)
+a = Dict.foldl (\\_ a s -> f a s) init
+"""
+                        ]
+        , test "should replace List.foldl (\\a s -> f a s) init (Dict.keys dict) by Dict.foldl (\\a _ s -> f a s) init dict" <|
+            \() ->
+                """module A exposing (..)
+a = List.foldl (\\a s -> f a s) init (Dict.keys dict)
```
jfmengels: Should we support/have tests for when the there are only some of the lambda parameters, like `\a -> f a`?

Issue #335, opened 2025-01-14
Further Tuple.mapFirst / Tuple.mapSecond simplification


jfmengels:
We currently do the following simplifications:

```elm
Tuple.first (Tuple.mapBoth changeFirst changeSecond tuple)
--> Tuple.first (Tuple.mapFirst changeFirst tuple)

Tuple.second (Tuple.mapBoth changeFirst changeSecond tuple)
--> Tuple.second (Tuple.mapSecond changeSecond tuple)
```

What I'm thinking is that we could probably simplify them even further:

```elm
Tuple.first (Tuple.mapFirst changeFirst tuple)
--> changeFirst (Tuple.first tuple)
```

And the same for `Tuple.second`.
This is a lot simpler, and if the tuple is a literal will be simplifier even further (`changeFirst (Tuple.first ( a, b ) )` -> `changeFirst a`).

The only thing I'm unsure about is when the map function is a lambda, because we'd end up with code like `(\x -> fn x y) (Tuple.first tuple)` which doesn't feel great. But I think that's probably okay.

PR #392, opened 2025-12-25
Add tuple access after mapping that part simplifications

lue-bird:
Implements/closes https://github.com/jfmengels/elm-review-simplify/issues/335
```elm
Tuple.first (Tuple.mapFirst changeFirst tuple)
--> changeFirst (Tuple.first tuple)

Tuple.second (Tuple.mapSecond changeSecond tuple)
--> changeSecond (Tuple.second tuple)
```
including composition.

Also changes the fix of
```elm
-- same for second
Tuple.first (Tuple.mapBoth changeFirst changeSecond tuple)
--> changeFirst (Tuple.first tuple)
-- instead of Tuple.first (Tuple.mapFirst changeFirst tuple)
```

The fixes are done in a let's say "creative" way to avoid extracting source code :sweat_smile:


jfmengels:
Looks great! Nice work on the creative fixes too :blush:

PR #394, opened 2025-12-26
Add List.foldl on List.reverse checks

jfmengels:
Adds this check from #2 (which we should at this point close and re-open in separate issues to make it easier to close them).

```elm
List.foldl f initial (List.reverse list)
--> List.foldr f initial list

List.foldr f initial (List.reverse list)
--> List.foldl f initial list
```

and composition


lue-bird:
Also: Not really necessary but if you want, add these to the correctness tests:
```elm
, Test.fuzz
  (Fuzz.list Fuzz.string)
  "List.reverse >> List.foldl is the same as List.foldr"
  (\list ->
      list
          |> List.reverse
          |> List.foldl (::) []
          |> Expect.equalLists
              (list
                  |> List.foldr (::) []
              )
  )
, Test.fuzz
  (Fuzz.list Fuzz.string)
  "List.reverse >> List.foldr is the same as List.foldl"
  (\list ->
      list
          |> List.reverse
          |> List.foldr (::) []
          |> Expect.equalLists
              (list
                  |> List.foldl (::) []
              )
  )
```

jfmengels:
Merged with your suggestions :+1:

Issue #317, opened 2024-08-24
`List.foldr (++) ""`


miniBill:
**What the rule should do:**

`List.foldr (++) "" => String.concat`

**Should this be part of the Simplify rule or should it be a new rule?**
Part of Simplify

**I am looking for:**
- Someone to implement it with/for me

PR #395, opened 2025-12-26
Add List.foldr (++) empty to concat simplifications

lue-bird:
```elm
List.foldr (++) "" list
--> String.concat list

List.foldr (++) [] list
--> List.concat list
```
Implementing the first simplification closes https://github.com/jfmengels/elm-review-simplify/issues/317
(Also when reviewing, the initial fix contains a bug that is later fixed)

Also fixes a rare bug when reducing lambdas


jfmengels:
Looks great, thanks!

PR #396, opened 2025-12-27
Optimize recursion in expressionContainsAnyVariableInIgnoring

jfmengels:


src/Simplify/AstHelpers.elm
```diff
@@ -662,127 +662,129 @@ getReducedLambda context expressionNode =
 
 expressionContainsAnyVariableInIgnoring : Set String -> Node Expression -> Node Expression -> Bool
 expressionContainsAnyVariableInIgnoring bindingsToCheckFor subToIgnore (Node expressionToCheckRange expressionToCheck) =
-    (Node.range subToIgnore /= expressionToCheckRange)
-        && (case expressionToCheck of
```
jfmengels: Since all recursive calls were under the `&&`, there wasn't actually any TCO.

src/Simplify/AstHelpers.elm
```diff
@@ -752,12 +752,8 @@ expressionContainsAnyVariableInIgnoring bindingsToCheckFor subToIgnore (Node exp
...
 
             Expression.CaseExpression caseOf ->
-                -- || TCO-ed
-                if expressionContainsAnyVariableInIgnoring bindingsToCheckFor subToIgnore caseOf.expression then
-                    True
-
-                else
-                    List.any
+                expressionContainsAnyVariableInIgnoring bindingsToCheckFor subToIgnore caseOf.expression
+                    || List.any
```
jfmengels: Since there isn't actually optimizable recursion in the `else` branch, we might as well use `||`.

lue-bird (approved):

src/Simplify/AstHelpers.elm
jfmengels: I agree, it probably doesn't matter too much 🤷‍♂️

PR #397, opened 2025-12-27
Fix removal of all arguments when simplifying lambda

jfmengels:
Bug fix: Simplifying directly applied lambdas now doesn't remove extra arguments. `(\_ a -> ...) b c` is now simplified to `(\a -> ...) c` instead of `(\a -> ...)`.


lue-bird (approved):
nice find

Issue #306, opened 2024-04-16
More List.minimum/maximum simplifications


jfmengels:
```elm
List.minimum [ 1, 2, 3 ]
--> Just 1
List.minimum (1 :: [ 2, 3 ])
--> Just 1

List.maximum [ 1, 2, 3 ]
--> Just 3

List.minimum [ a, 1 ]
-- Not reported, but if there are duplicate value, then we can remove some of them
List.minimum [ 3, a, 1 ]
--> List.minimum [ a, 1 ]
List.minimum (3 :: list ++ [5])
--> List.minimum (3 :: list)
List.minimum ([3] ++ list ++ [5])
--> List.minimum ([3] ++ list)

List.minimum (List.repeat <positiveInt> x)
--> Just x
List.maximum (List.repeat <positiveInt> x)
--> Just x
```

If `x <= y`

```elm
List.minimum (List.range x y)
--> Just x
List.maximum (List.range x y)
--> Just y
```

morteako:
Some more

```elm
-- known n
-- same for maximum
List.minimum (List.repeat n x) 
--> Just x -- if n > 0
--> Nothing -- if n <= 0

-- unknown n
-- same for maximum
List.minimum (List.repeat n x) 
--> if n > 0 then Just x else Nothing


-- same for maximum
List.minimum (List.reverse list)
--> List.minimum list


-- same for maximum, sortBy f, sortWith f
List.minimum (List.sort list)
--> List.minimum sort
```

jfmengels:
Nice 👍 

> List.minimum (List.repeat n x) 
--> Nothing -- if n <= 0

We already have this one indirectly, because `List.repeat n x` will already be replaced by `[]` (and then the `List.minimum [] -> Nothing` will kick in.

> -- same for maximum, sortBy f, sortWith f
List.minimum (List.sort list)
--> List.minimum sort

I think you meant `List.minimum list` as the result.

PR #399, opened 2025-12-27
More List.range simplifications

lue-bird:
A few very simple ones
```elm
List.minimum (List.range 2 3)
--> Just 2

List.maximum (List.range 2 3)
--> Just 3

List.range n n
--> [ n ]
```
The first 2 are mentioned in https://github.com/jfmengels/elm-review-simplify/issues/306, last one is listed in https://github.com/jfmengels/elm-review-simplify/issues/2


jfmengels:
Very nice :+1: 
I'll update the issues

PR #401, opened 2025-12-29
More fold to list simplifications

lue-bird:
```elm
List.foldr (::) [] list
--> list

List.foldl (::) [] list
--> List.reverse list

Set.foldr (::) [] set
--> Set.toList set

-- same for foldr, when `expectNaN` is not enabled
Set.foldl Set.insert Set.empty set
--> set

Dict.foldr (\k _ ks -> k :: ks) [] dict
--> Dict.keys dict

Dict.foldr (\_ v vs -> v :: vs) [] dict
--> Dict.values dict

-- same for foldr, when `expectNaN` is not enabled
Dict.foldl Dict.insert Dict.empty dict
--> dict

Dict.foldr (\k v list -> ( k, v ) :: list) Dict.empty dict
--> Dict.toList dict
```

Includes some nice improvements to normalize.

There are so many more simplifications of this kind (filter, map, fromList, ..., the append ones with flipped arguments, even combining into non-empty accumulator as union, ++ etc). I'll probably try to do something more useful next :sweat_smile:

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -9304,7 +9304,7 @@ dictFoldrChecks =
                                             )
 
                                     else if
-                                        case getFunctionWithFirstSecondAndIncomingAsTuple checkInfo checkInfo.firstArg of
+                                        case getFunctionWithFirstAndSecondIncomingAsTuple checkInfo checkInfo.firstArg of
                                             Nothing ->
                                                 False
 
@@ -9330,11 +9330,11 @@ dictFoldrChecks =
         )
 
 
-getFunctionWithFirstSecondAndIncomingAsTuple :
+getFunctionWithFirstAndSecondIncomingAsTuple :
     AstHelpers.ReduceLambdaResources a
     -> Node Expression
     -> Maybe (Node Expression)
-getFunctionWithFirstSecondAndIncomingAsTuple resources expressionNode =
+getFunctionWithFirstAndSecondIncomingAsTuple resources expressionNode =
     case AstHelpers.getCollapsedLambda expressionNode of
         Nothing ->
             Nothing


src/Simplify.elm
```diff
@@ -8757,7 +8760,32 @@ setFoldlChecks =
...
+                |> onNothing
+                    (\() ->
+                        case secondArg checkInfo of
+                            Nothing ->
+                                Nothing
+
+                            Just initialArg ->
+                                if
+                                    isEmptyList initialArg
+                                        && AstHelpers.isSpecificUnappliedBinaryOperation "::" checkInfo checkInfo.firstArg
```
jfmengels: Does this also detect `\a b -> a :: b`?

src/Simplify.elm
```diff
@@ -8788,6 +8791,36 @@ setFoldrChecks =
         )
 
 
+setFoldChecks : CallCheckInfo -> Maybe (Error {})
+setFoldChecks checkInfo =
+    emptiableFoldChecks setCollection checkInfo
+        |> onNothing
+            (\() ->
+                if checkInfo.expectNaN then
```
jfmengels:
```elm
a = Set.fromList [NaN,NaN,NaN]

Set.foldl Set.insert Set.empty a
-->
Set.fromList [NaN,NaN,NaN]
```

These are the same, though it now fails a `a == <result>` check. I think all cases where we can make `==` return `True` instead is better, but maybe that is just a biased opinion. I'm missing use-cases where having `==` return `False` in this sort of case is useful :man_shrugging: 

Maybe let's keep it the way you wrote it, that's "safer", but yeah... :man_shrugging:

CHANGELOG.md
```diff
@@ -106,6 +106,7 @@ Other improvements:
   to for example fix `(if c then 2 else 3) == (if c then 1 else 4)` to `False`
 - Now evaluates `<`, `<=`, `>=`, `>` for any two comparable operands to for example fix `"a" < "b"` to `True`
 - Now fixes `Tuple.first (Tuple.mapBoth changeFirst changeSecond tuple)` to `changeFirst (Tuple.first tuple)` instead of `Tuple.first (Tuple.mapFirst changeFirst tuple)` (same for second)
+- Now recognizes more lambdas as equivalent, to for example detect equal branches like `if c then f else \a -> f a`
```
jfmengels: Damn, that's nice! :heart:

jfmengels:
Beautiful work here @lue-bird, thank you so much for this :heart:

Issue #112, opened 2023-06-24
Simplify to a more optimized mapping operations


jfmengels:
**Example of things the rule would report:**

```elm
list
  |> List.map f
  |> List.head
-->
list
  |> List.head
  |> Maybe.map f

List.head (List.map f list)
-->
Maybe.map f (List.head list)
```

I'm guessing we can apply this idea to other patterns as well. For instance for tuples:

```
tuple
  |> Tuple.mapFirst f
  |> Tuple.first
-->
tuple
  |> Tuple.first
  |> f
-- and similarly for second
```

For tuples, we could even notice unnecessary operations
```
tuple
  |> Tuple.mapSecond f
  |> Tuple.first
-->
tuple
  |> Tuple.first
-- and similarly for second
```

If you can think of other operations where mapping (or similar operations) can be improved, please mention it!

**I am looking for:**

- Feedback
- Someone to implement it

PR #405, opened 2026-01-03
Add reorder operations for performance simplifications

lue-bird:
Implements/closes https://github.com/jfmengels/elm-review-simplify/issues/112 and some similar operations
```elm
List.head (List.map f list)
--> Maybe.map f (List.head list)

List.take n (List.map f list)
--> List.map f (List.take n list)

List.take n (List.indexedMap f list)
--> List.indexedMap f (List.take n list)

List.drop n (List.map f list)
--> List.map f (List.drop n list)

Array.get i (Array.map f array)
--> Maybe.map f (Array.get i array)

Array.slice start end (Array.map f array)
--> Array.map f (Array.slice start end array)

Array.slice 0 end (Array.indexedMap f array)
--> Array.indexedMap f (Array.slice 0 end array)

Dict.remove k (Dict.map g dict)
--> Dict.map g (Dict.remove k dict)

Dict.filter (\k _ -> f k) (Dict.map g dict)
--> Dict.map g (Dict.filter (\k _ -> f k) dict)

Dict.diff (Dict.map f dict) remove
--> Dict.map f (Dict.diff dict remove)
```
including composition where possible.

#### skipped
```elm
Array.get i (Array.indexedMap f array)
--> Maybe.map (f i) (Array.get i array)
```
due to the duplicated index expression. The resulting expression is faster most of the time but potentially less simple.
And for the same reason similar operations are skipped like
```elm
List.drop n (List.indexedMap f list)
--> List.indexedMap ((+) (max 0 n) >> f) (List.drop n)

Array.slice start end (Array.indexedMap f array)
--> Array.indexedMap ((+) (max 0 start) >> f) (Array.slice start end array)
```

#### not sure
```elm
List.head (List.indexedMap f list)
--> Maybe.map (f 0) (List.head list)
```
If deemed useful, I'll implement it in a separate PR

----

Was feeling a bit stuck while completing this PR but have some new ideas for future PRs now

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -7124,7 +7124,8 @@ listHeadChecks =
             , combinedFn = Fn.List.minimum
             }
         , earlierOperationCanBeMovedAfterAsForPerformanceChecks
-            { earlierFn = Fn.List.map
+            { laterOperationArgsDescription = Nothing
+            , earlierFn = Fn.List.map
             , earlierFnArgCount = 2
             , earlierFnOperationArgsDescription = "function"
             , asLaterFn = Fn.Maybe.map
@@ -7152,7 +7153,8 @@ or with `{ earlierFn = Fn.List.take, asLaterFn = Fn.Maybe.map }`:
 
 -}
 earlierOperationCanBeMovedAfterAsForPerformanceChecks :
-    { earlierFn : ( ModuleName, String )
+    { laterOperationArgsDescription : Maybe String
+    , earlierFn : ( ModuleName, String )
     , earlierFnArgCount : Int
     , earlierFnOperationArgsDescription : String
     , asLaterFn : ( ModuleName, String )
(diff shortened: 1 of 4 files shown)


jfmengels:
This is beautiful @lue-bird :blue_heart: 
Thank you so much!

> skipped
```elm
Array.get i (Array.indexedMap f array)
--> Maybe.map (f i) (Array.get i array)
```
Yeah... let's skip this. If `i` is a large expression, this might look terrible. But even if it's a simple number, it means there are now 2 places where you have to update that number if you need to change that code `Maybe.map (f 2) (Array.get 2 array)`. I'm thinking, it might be okay if it's a simple variable `Maybe.map (f someVar) (Array.get someVar array)`, but then it becomes weird that this simplification only applies in specific cases. So yeah, let's skip this I think :man_shrugging: 

> And for the same reason similar operations are skipped like
```elm
List.drop n (List.indexedMap f list)
--> List.indexedMap ((+) (max 0 n) >> f) (List.drop n)

Array.slice start end (Array.indexedMap f array)
--> Array.indexedMap ((+) (max 0 start) >> f) (Array.slice start end array)
```
Yeah these look pretty terrible, definitely does not feel like a simplification. Eventually we could report it without a fix, but this would be an exception...

> not sure
```elm
List.head (List.indexedMap f list)
--> Maybe.map (f 0) (List.head list)
```
I think this one makes a lot of sense :+1:

PR #409, opened 2026-01-06
More compare element count against 0/1 simplifications

lue-bird:
Implements the remaining pieces of and closes https://github.com/jfmengels/elm-review-simplify/issues/26 including a bunch of similar simplifications
```elm
-- The following simplifications work with all element count fns: Set.size, Dict.size, List.length, Array.length

List.length l < 1 --> List.isEmpty l
0 < List.length l --> not (List.isEmpty l)
List.length l <= 0 --> List.isEmpty l
1 <= List.length l --> not (List.isEmpty l)
0 >= List.length l --> List.isEmpty l
List.length l >= 1 --> not (List.isEmpty l)
1 > List.length l --> List.isEmpty l
List.length l > 0 --> not (List.isEmpty l)


-- same with all the compare operators with a 0/1 on the left
(>=) 0 << List.length
--> List.isEmpty

case Set.size set of
    0 -> x
    _ -> y
--> if Set.isEmpty set then x else y
```

There is a commented-out line that would also add these simplifications for `String.length` to `String.isEmpty` (commented out because it argues `== ""` might be simpler). I would personally argue that `String.isEmpty` is a perfectly fine fix, thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+The rule now simplifies:
+- `String.length str == 0` to `String.isEmpty str` (including all the variations mentioned in "Other improvements"ü)
+
 ## [Unreleased]
 
 Other improvements:
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -444,6 +444,9 @@ Destructuring using case expressions
     String.length "abc"
     --> 3
 
+    String.length str == 0
+    --> String.isEmpty str
+
     String.repeat n ""
     --> ""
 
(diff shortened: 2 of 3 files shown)


jfmengels:
Very nice changes :blush: 

> There is a commented-out line that would also add these simplifications for String.length to String.isEmpty (commented out because it argues == "" might be simpler). I would personally argue that String.isEmpty is a perfectly fine fix, thoughts?

What I think would be borderline is changing `x == ""` to `String.isEmpty x` (at the former is for some proably easier to read), but changing `String.length` + comparison to `String.isEmpty` is perfectly fine I believe alos :+1: 

(Feel free to merge if you're content with the state of the PR)

Issue #404, opened 2026-01-01
Additional unnecessary on specific fn simplifications


lue-bird:
For all operations below, the earlier operation is unnecessary and can be removed:
```elm
List.head << List.intersperse sep
Result.toMaybe << Result.mapErr f
Dict.diff dict << Dict.map f
List.sum << List.reverse -- same with any sort
List.product << List.reverse -- same with any sort
List.minimum << List.reverse -- same with any sort, when expectNaN is not enabled
List.maximum << List.reverse -- same with any sort, when expectNaN is not enabled
Set.fromList << List.reverse -- same with any sort, when expectNaN is not enabled
```
Since we want to completely avoid false positives, here's a tricky one we need to skip:
```elm
List.any f << List.reverse -- same with any sort
List.all f << List.reverse -- same with any sort
```
since in theory `f` could be so expensive that shortcutting as early as possible would be more important than avoiding the extra list traversal and allocation (realistically, this is maybe a 0.01% case)

PR #412, opened 2026-01-07
Add more "unnecessary specific fns before" simplifications

lue-bird:
Implements exactly the simplifications listed in https://github.com/jfmengels/elm-review-simplify/issues/404


jfmengels:
Thank you 😊

Issue #297, opened 2024-03-07
filter, then isEmpty is equivalent to not any


lue-bird:
```elm
List.isEmpty (List.filter isOdd list)
--> not (List.any isOdd list)
```
(strangely, no other published elm/... module exposes both any and filter.)

If we want to save an extra step, we can also introduce separate checks for the negated applications like
```elm
not (List.isEmpty (List.filter isOdd list))
--> List.any isOdd list
```

(found in the [FSharpLint default config](https://github.com/fsprojects/FSharpLint/blob/0007ec1e83a1043d74bcd5a0067c6483657faa26/src/FSharpLint.Core/DefaultConfiguration.json#L320))

PR #422, opened 2026-01-09
Add List.isEmpty on List.filter simplifications

lue-bird:
Implements/closes https://github.com/jfmengels/elm-review-simplify/issues/297
```elm
List.isEmpty (List.filter f list)
--> not (List.any f list)

List.isEmpty (List.filter (not << f) list)
--> List.all f list
```
including composition and anything recognizable as an "into Basics.not" function


jfmengels:
Very nice! Now I'm sad that `Set/Dict/Array` don't have `all` nor `any` :smiling_face_with_tear:

PR #425, opened 2026-01-12
Make determining collection size more powerful and use it while determining number bounds

lue-bird:
To enable simplifications like
```elm
(String.length (String.cons h t) + List.length (x :: xs)) >= 2 --> True
String.length (String.fromList [ c0, c1 ] <= 4 --> True
```
Also contains a bug fix for `String.cons` only adding 1 to the collection size (added char could be 2 wide).

The commits are a bit confusing as I changed my mind about the final direction a few times.
For example, the commit [make AST helpers work on normalized syntax](https://github.com/jfmengels/elm-review-simplify/commit/b3e77a2b5d204334a8165093e4ad91bd458c25ff) ended up being totally unnecessary in the end, and we can revert it if you think it makes more sense.

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -10941,7 +10941,7 @@ dictMemberChecks =
                                 , elements =
                                     { countDescription = "size"
                                     , elementDescription = "key"
-                                    , determineCount = dictDetermineSize
+                                    , determineCount = determineCollectionSize
                                     , get = dictGetKeys
                                     }
                                 }
@@ -12614,7 +12614,7 @@ listCollection =
         { get = listGetElements
         , elementDescription = "element"
         , countDescription = "length"
-        , determineCount = listDetermineLength
+        , determineCount = determineCollectionSize
         }
     , wrap = listSingletonConstruct
     , mapFn = Fn.List.map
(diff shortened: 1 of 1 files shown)


src/Simplify.elm
```diff
@@ -18554,43 +18621,60 @@ collectionInsertChecks collection =
...
+                            (String.fromInt collectionSize.min)
+                        ]
+                    )
+
+            else
+                Nothing
+
 
 {-| On a "take" operation that returns a given number of elements _from the either side_
 (for example `String.right` or `List.take`).
```
jfmengels: Functions like `List.take` and `List.drop` (same for String) should also be able to influence the max size of collections. AFAICT (to the extent I've reviewed at least) I can't see them affecting the collection size.

src/Simplify.elm
```diff
@@ -18975,6 +18698,350 @@ collectionSizeToNumberBounds collectionSize =
...
+                            (normalListDetermineLength leftArg)
+                            (normalListDetermineLength rightArg)
+
+                    _ ->
+                        collectionSizeUnknown
+          )
+
+        -- Array
+        , ( Fn.Array.empty, \_ -> collectionSizeExact 0 )
+        , ( Fn.Array.fromList
```
jfmengels: I think we can also `map` functions which don't change the size of the collection (except for `Set.map` which can reduce it)

Issue #427, opened 2026-01-13
not << List.any not --> List.all identity


jfmengels:
```elm
not << List.any not
--> List.all identity

not << List.all not
--> List.any identity
```

I discovered the former one because of the following simplifications we already do
```elm
x
  |> List.filter not
  |> List.isEmpty
-->
x
  |> List.any not
  |> not
```

Confirmed by:
```elm
        , Test.fuzz
            (Fuzz.list Fuzz.bool)
            "not << List.any not is the same as List.all identity"
            (\list ->
                (not <| List.any not list)
                    |> Expect.equal
                        (List.all identity list)
            )
        , Test.fuzz
            (Fuzz.list Fuzz.bool)
            "not << List.all not is the same as List.any identity"
            (\list ->
                (not <| List.all not list)
                    |> Expect.equal
                        (List.any identity list)
            )
```

Probably good to do relatively close to https://github.com/jfmengels/elm-review-simplify/issues/424 since they deal with the same functions.

PR #429, opened 2026-01-13
Add Basics.not on List.any/all with a function into Basics.not simplifications

lue-bird:
Implements/closes https://github.com/jfmengels/elm-review-simplify/issues/427:
```elm
Basics.not (List.all (not << f) list)
--> List.any f list

Basics.not (List.all not list)
--> List.any identity list

Basics.not (List.any (not << f) list)
--> List.all f list

Basics.not (List.any not list)
--> List.all identity list
```
including composition

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1033,7 +1033,7 @@ Destructuring using case expressions
     List.all f (List.repeat n a)
     --> n <= 0 || f a
 
-    Basics.not (List.all (not << f) list)
+    not (List.all (not << f) list)
     --> List.any f list
 
     List.any f []
@@ -1060,7 +1060,7 @@ Destructuring using case expressions
     List.any f (List.repeat n a)
     --> n >= 1 && f a
 
-    Basics.not (List.any (not << f) list)
+    not (List.any (not << f) list)
     --> List.all f list
 
     List.range 6 3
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -9746,44 +9734,50 @@ listIsEmptyOnListFilterChecks =
     }
 
 
-getFunctionIntoNot : AstHelpers.ReduceLambdaResources a -> Node Expression -> Maybe (List { withNotRange : Range, withoutNot : Node Expression })
+getFunctionIntoNot : AstHelpers.ReduceLambdaResources (QualifyResources a) -> Node Expression -> Maybe (List Fix)
```
jfmengels: Now that `getFunctionIntoNot` is more powerful, I wonder whether we could also check for `==` replaceable by `/=` (and the other way around (maybe only without `expectNaN`) and same for `>=` like operators.

That way, we could simplify `not <| List.any (\a -> a == b) list` by `List.all (\a /= b) list`.

src/Simplify.elm
```diff
@@ -1057,6 +1060,9 @@ Destructuring using case expressions
     List.any f (List.repeat n a)
     --> n >= 1 && f a
 
+    Basics.not (List.any (not << f) list)
```
jfmengels: Same here
```suggestion
    not (List.any (not << f) list)
```

Issue #424, opened 2026-01-10
simplify any/all to isEmpty


lue-bird:
For `String` and `List`:
```elm
any (always True) --> not << isEmpty
all (always False) --> isEmpty
```

PR #435, opened 2026-01-14
Add any/all (always True/False) simplifications

lue-bird:
Implements/closes https://github.com/jfmengels/elm-review-simplify/issues/424
```elm
String.any (always True) str
--> not (String.isEmpty str)

String.all (always False) str
--> String.isEmpty str

List.all (always False) list
--> List.isEmpty list

List.any (always True) list
--> not (List.isEmpty list)
```
including without the last argument


jfmengels:
:tada:

Issue #432, opened 2026-01-13
Remove unnecessary operations when checking collection size


jfmengels:
```elm
-- Same for Array, Dict but not Set, String
List.length (List.map f list)

-- Same for Array, Dict, Set, String
List.isEmpty (List.map f list)
--> List.isEmpty list

-- Same for List.sort* functions
List.length/isEmpty (List.reverse list)
--> List.length/isEmpty list

-- Same for Array
List.length/isEmpty (List.indexedMap f list)
--> List.length/isEmpty list

-- ✅
List.length/isEmpty (Array.toIndexedList array)
--> Array.length/isEmpty array
```

jfmengels:
I think there's a class of collection functions that don't change the size of the collection, so it can be interesting to just a list/collection of those, and remove them when they're chained into a length/emptiness check.

Here are some more (only writing the example with size functions):
```elm
List.length (List.sort list) -- also sortBy and sortWith
--> List.length list

Array.length (Array.map f array)
--> Array.length array

Array.length (Array.set index a array)
--> Array.length array

Array.size (Array.indexedMap f array)
--> Array.length array

Dict.size (Dict.map f dict)
--> Dict.size dict

String.length (String.reverse str)
--> String.length str
```

Issue #433, opened 2026-01-13
Account functions when computing collection size


jfmengels:
I don't think these should be simplified, but they could be accounted for in the collection size computation

```elm
-- Same for map3, ...
List.map2 f listA listB
  |> List.length
-- Consider the length as
-- { min = min (List.length listA) (List.length listB)
-- , max = <if both are known, min of the 2 maxes. Otherwise whichever is not Nothing>
-- }

List.intersperse x list
-- Consider the length as
-- { min = 2 * (<previous min> - 1) + 1
-- , max = Maybe.map (\length ->  2 * (length - 1) + 1) <previous max>
-- }

-- Same for Dict
Set.union a b
  |> Set.size
-- Consider the size as
-- { min = max (Set.size a).min (Set.size b).min
-- , max = (Set.size a).max + (Set.size b).max
-- }

-- Same for Dict
Set.intersect a b
  |> Set.size
-- Consider the size as
-- { min = 0
-- , max = min (Set.size a).max (Set.size b).max
-- }
```

lue-bird:
same with
```elm
Set/Dict.insert k [v] collection --> 0→1 + collectionSize but importantly: if min was 0 before, set it to 1
Set/Dict.remove k collection --> -1→0 + collectionSize
String.lines/String.words list --> each (max 1) listSize
```

PR #439, opened 2026-01-15
Unnecessary fn before is empty and length

lue-bird:
Implements/closes https://github.com/jfmengels/elm-review-simplify/issues/432#issuecomment-3755468611
```elm
-- same for String.reverse
String.isEmpty (String.map f str)
--> String.isEmpty str

String.length (String.reverse str)
--> String.length str

-- same for List.sort, List.sortBy, List.sortWith, List.map, List.indexedMap
List.isEmpty (List.reverse list)
--> List.isEmpty list

-- same for List.sort, List.sortBy, List.sortWith, List.map, List.indexedMap
List.length (List.reverse list)
--> List.length list

-- same for Array.indexedMap
Array.isEmpty (Array.map array)
--> Array.isEmpty array

-- same for Array.indexedMap
Array.length (Array.map array)
--> Array.length array

Set.isEmpty (Set.map f set)
--> Set.isEmpty set

Dict.isEmpty (Dict.map f dict)
--> Dict.isEmpty dict

Dict.size (Dict.map f dict)
--> Dict.size dict

-- below were not listed in the issue

String.isEmpty (String.fromList list)
--> List.isEmpty list

List.isEmpty (String.toList string)
--> String.isEmpty string
```
including composition.
No changes have been made to `determineCollectionSize`, I'll probably batch those with https://github.com/jfmengels/elm-review-simplify/issues/433

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -1255,7 +1255,7 @@ Destructuring using case expressions
     --> True
 
     -- same for Array.indexedMap
-    Array.isEmpty (Array.map array)
+    Array.isEmpty (Array.map f array)
     --> Array.isEmpty array
 
     Array.repeat 0 x
@@ -1283,7 +1283,7 @@ Destructuring using case expressions
     --> max 0 n
 
     -- same for Array.indexedMap
-    Array.length (Array.map array)
+    Array.length (Array.map f array)
     --> Array.length array
 
     Array.filter f (Array.filter f array)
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -1233,6 +1248,10 @@ Destructuring using case expressions
     Array.isEmpty Array.empty
     --> True
 
+    -- same for Array.indexedMap
+    Array.isEmpty (Array.map array)
```
jfmengels: Should be `Array.map f array`, same in the next diff and for changelog

PR #441, opened 2026-01-17
Expand determining collection size

lue-bird:
Implements/closes https://github.com/jfmengels/elm-review-simplify/issues/433 including the unnecessary fns before from https://github.com/jfmengels/elm-review-simplify/pull/439.
It covers: List.intersperse/sort*/reverse/map/indexedMap/map2-5, String.toList/lines/words/reverse/map, Array.to*List/map/indexedMap, Set.toList/map/insert/remove/union/intersect, Dict.toList/map/insert/remove/union/intersect, `if then else`, `let in`, `case of`

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -19359,58 +19359,22 @@ normalFnOrFnCallDetermineCollectionSizeDict =
                         collectionSizeUnknown
           )
         , ( Fn.List.sort
-          , \args ->
-                case args of
-                    [ Node _ listArg ] ->
-                        normalDetermineCollectionSize listArg
-
-                    _ ->
-                        collectionSizeUnknown
+          , normalArgsDetermineCollectionSizeAtIndex 0
           )
         , ( Fn.List.sortBy
-          , \args ->
-                case args of
-                    [ _, Node _ listArg ] ->
-                        normalDetermineCollectionSize listArg
-
-                    _ ->
-                        collectionSizeUnknown
+          , normalArgsDetermineCollectionSizeAtIndex 1
           )
         , ( Fn.List.sortWith
-          , \args ->
-                case args of
-                    [ _, Node _ listArg ] ->
-                        normalDetermineCollectionSize listArg
-
-                    _ ->
-                        collectionSizeUnknown
+          , normalArgsDetermineCollectionSizeAtIndex 1
           )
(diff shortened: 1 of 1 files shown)


tests/Simplify/ComparisonTest.elm
```diff
@@ -1310,6 +1310,48 @@ a = Set.size (Set.union (Set.fromList (List.take 2 xs)) (List.take 1 ys)) <= 3
                             |> Review.Test.whenFixed """module A exposing (..)
 import Set
 a = True
+"""
+                        ]
+        , test "should replace Set.size (Set.intersect (Set.fromList (List.take 2 xs)) (List.take 3 ys)) <= 2 by True" <|
+            \() ->
+                """module A exposing (..)
+import Set
+a = Set.size (Set.intersect (Set.fromList (List.take 2 xs)) (List.take 3 ys)) <= 2
```
jfmengels: This doesn't compile because the second argument to Set.intersect is a List and not a Set. Same for the next test

src/Simplify.elm
```diff
@@ -19402,6 +19456,58 @@ normalFnOrFnCallDetermineCollectionSizeDict =
...
+                        collectionSizeUnknown
+          )
+        , ( Fn.Dict.toList
+          , \args ->
+                case args of
+                    [ Node _ dictArg ] ->
+                        normalDetermineCollectionSize dictArg
+
+                    _ ->
+                        collectionSizeUnknown
```
jfmengels: These definitions are quite repetitive, I think it would make sense to have a helper function like this:

```elm
sizeIsTheSameAsArgAtPosition : Int -> List (Node Expression) -> CollectionSize
sizeIsTheSameAsArgAtPosition collectionIndex =
  \args ->
    case List.drop collectionIndex args of
      [ Node _ collection ] -> normalDetermineCollectionSize  collection
      _ -> collectionSizeUnknown
    

---

[ ...
, ( Fn.List.indexedMap, sizeIsTheSameAsArgAtPosition 1 )
, ...
, ( Fn.String.toList, sizeIsTheSameAsArgAtPosition 0 )
, ...
]
```

lue-bird: Good suggestion. Didn't add this before because I wasn't able to come up with a name.

Issue #327, opened 2024-12-18
`Dict.filter (\k _ -> k /= foo)` => `Dict.remove foo`


miniBill:
**What the rule should do:**
`Dict.filter (\k _ -> k /= foo)` => `Dict.remove foo`

**What problems does it solve:**
Change from `O(n)` to `O(log n)` complexity, and reduce memory usage.

**Should this be part of the Simplify rule or should it be a new rule?**
Part of `Simplify`

**I am looking for:**
- Someone to implement it with/for me

lue-bird:
also
```elm
Set.filter (\k -> k /= needle) set
--> Set.remove needle set
```

lue-bird:
Yeah, NaNs do not follow this rule
```elm
Set.fromList [ 0/0, 0/0 ]
--> Set.fromList [NaN,NaN]
```
so we should only do it if `expectNaN` is configured False.

PR #443, opened 2026-01-18
Add Dict/Set.filter to remove siplifications

lue-bird:
Implements/closes https://github.com/jfmengels/elm-review-simplify/issues/327
```elm
-- both when `expectNaN` is not enabled
Set.filter (\k -> k /= specificKey) set
--> Set.remove specificKey set

Dict.filter (\k _ -> k /= specificKey) dict
--> Dict.remove specificKey dict
```
including composition, prefix operator application, Basics.always etc

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -10983,7 +10983,7 @@ setFilterChecks =
                                     , details =
                                         [ "You can replace this call by "
                                             ++ qualifiedToString Fn.Set.remove
-                                            ++ " with the specific value you compared against which meant for this exact purpose and will also be faster."
+                                            ++ " with the specific value you compared against which is meant for this exact purpose and will also be faster."
                                         ]
                                     }
                                     checkInfo.fnRange
@@ -11664,7 +11664,7 @@ dictFilterChecks =
                                             , details =
                                                 [ "You can replace this call by "
                                                     ++ qualifiedToString Fn.Dict.remove
-                                                    ++ " with the specific value you compared against which meant for this exact purpose and will also be faster."
+                                                    ++ " with the specific value you compared against which is meant for this exact purpose and will also be faster."
                                                 ]
                                             }
                                             checkInfo.fnRange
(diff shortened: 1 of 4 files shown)


src/Simplify.elm
```diff
@@ -15523,27 +15596,31 @@ operationWithEqualsConstantIsEquivalentToFnWithThatConstantCheck replacementFn c
...
                     case Node.value (AstHelpers.removeParens lambda.expression) of
-                        Expression.OperatorApplication "==" _ left right ->
-                            let
-                                nodeToFind : Expression
-                                nodeToFind =
-                                    Expression.FunctionOrValue [] var
-                            in
-                            if (Node.value left == nodeToFind) && not (expressionContainsVariable var right) then
-                                Just { constant = right }
+                        Expression.OperatorApplication operator _ left right ->
```
jfmengels: We could destructure left and right here.

CHANGELOG.md
```diff
@@ -76,6 +76,8 @@ The rule now simplifies:
 - `String.isEmpty (String.fromList list)` to `List.isEmpty list`
 - `List.filter f (List.reverse list)` to `List.reverse (List.filter f list)` (same for `List.sort`, `List.sortBy` and `List.sortWith`)
 - `String.filter f (String.reverse str)` to `String.reverse (String.filter f str)`
+- `Set.filter (\k -> k /= specificKey) set` to `Set.remove specificKey set` when [`expectNaN`] is not enabled
+- `Dict.filter (\k _ -> k /= specificKey) dict` to `Dict.remove specificKey dict` when [`expectNaN`] is not enabled
```
jfmengels: We could try and detect whether `specificKey` is a String (or other non-NaN compatible value?), and in that case make that work even if `expectNaN` is enabled.

jfmengels: I don't know whether it makes sense to do basic type inference (just by looking at type annotations).

For instance, if we see
```elm
let
  specificKey : String
  specificKey = ...
in
Dict.filter (\k _ -> k /= specificKey) dict
```
then we can apply it regardless of NaN.

Not necessarily for this PR obviously, and I'm thinking we could apply this idea in a lot more places. If you think that's viable, let's create an issue.

tests/Simplify/DictTest.elm
```diff
@@ -1750,6 +1750,40 @@ a = Dict.map g >> Dict.filter (\\k _ -> f k)
...
+        , test "should replace Dict.filter (\\k _ -> k /= (f <| x)) dict by always Dict.remove (f <| x) dict" <|
+            \() ->
+                """module A exposing (..)
+import Dict
+a = Dict.filter (\\k _ -> k /= (f <| x)) dict
+"""
+                    |> whenNotExpectingNaN Review.Test.run
+                        [ Review.Test.error
+                            { message = "Dict.filter checking each key for inequality with a specific value is the same as Dict.remove"
+                            , details = [ "You can replace this call by Dict.remove with the specific value you compared against which meant for this exact purpose and will also be faster." ]
```
jfmengels: "which IS meant"

PR #444, opened 2026-01-18
Simplify identity lambda applications

### LINKED ISSUE #421: Simplify application of identity lambda
We currently simplify this

```elm
identity a
--> a
```

but we don't do it for
```elm
(\b -> b) a
--> a
```
Consider how we do identity simplifications everywhere else, I think this would be a reasonable addition.

jfmengels:
Closes #421 

- `(\x -> x) data` to `data`
  - We previously supported `identity data` -> `data` but not with lambdas
  - I changed the wording of the error message as well. Maybe we should just always say "the identity function", I'd be happy with that as well.
- Enabled simplifications for lambdas that were disabled when the argument was itself a pipeline
  - `(\_ -> x) <| f <| y` to `x`
  - `(\() -> x) <| f <| y` to `x`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 
 The rule now simplifies:
 - `(\x -> x) data` to `data`
+- `(\_ -> x) <| f <| y` to `x` (previously not always applied in pipelines)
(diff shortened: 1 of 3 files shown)


jfmengels:
Good idea, I've created #452 from your comment.

Issue #449, opened 2026-01-18
Simplify List.map Tuple.first (Array.toIndexedList array)


jfmengels:
```elm
List.map Tuple.first (Array.toIndexedList array)
--> List.range 0 (Array.length array - 1)
```

lue-bird:
`Array.length array - 1`

PR #454, opened 2026-01-18
Add List.map Tuple.first on Array.toIndexedMap simplifications

lue-bird:
Implements/closes https://github.com/jfmengels/elm-review-simplify/issues/449
```elm
List.map Tuple.first (Array.toIndexedList array)
--> List.range 0 (Array.length array - 1)
```
I'm not actually sure this is simpler but it's definitely faster

Also, I have an implementation for
```elm
List.map Tuple.first << Array.toIndexedList
--> List.range 0 << (+) -1 << Array.length
```
Do you think this should be added?

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -8776,8 +8776,8 @@ listMapChecks =
 
 listMapTupleFirstOnArrayToIndexedListCheck : IntoFnCheck
 listMapTupleFirstOnArrayToIndexedListCheck =
-    intoFnCheckOnlyCall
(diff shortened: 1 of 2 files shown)


jfmengels (approved):

jfmengels:
> I'm not actually sure this is simpler but it's definitely faster

I think it's simpler. We immediately see we're just constructing a range, whereas before you would have had to figure that out.

> Also, I have an implementation for

Yes, let's do that 👍

Issue #455, opened 2026-01-19
Simplify duplicate conditionals


jfmengels:
We currently simplify
```elm
if condition then
    x
else if condition then
    y
else
    z
-->
if condition then
    x
else
    z
```
but we apparently don't do it for more complex operations like function calls (haven't delved deeper into what our apparent limit is), as this doesn't (but should) get simplified:
```elm
if condition arg then
    x
else if condition arg then
    y
else
    z
```

PR #458, opened 2026-01-19
Infer and Normalize improvements

lue-bird:
Makes normalize and infer more powerful, and closes https://github.com/jfmengels/elm-review-simplify/issues/455

To be honest, for `Infer` I still don't understand why it works for some more complex cases like nested `||` and `&&` but I'm not complaining :sweat_smile:

### DIFF
--- a/src/Simplify/Normalize.elm
+++ b/src/Simplify/Normalize.elm
@@ -428,15 +428,7 @@ addToFunctionCall resources functionCall extraArgument =
                 (Expression.RecordAccess extraArgument (Node.empty (String.dropLeft 1 fieldAccess)))
 
         Expression.FunctionOrValue [ "Basics" ] "not" ->
-            case Node.value extraArgument of
-                Expression.FunctionOrValue [ "Basics" ] "True" ->
-                    expressionFalse
-
-                Expression.FunctionOrValue [ "Basics" ] "False" ->
-                    expressionTrue
-
-                _ ->
-                    Expression.Application [ functionCall, extraArgument ]
+            createNot resources extraArgument
 
         Expression.FunctionOrValue [ "Basics" ] "negate" ->
             createNegation (Node.value extraArgument)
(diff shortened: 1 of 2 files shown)


src/Simplify/Normalize.elm
```diff
@@ -311,6 +294,18 @@ createOperation resources operator left right =
...
+        ">" ->
+            infer resources (Expression.OperatorApplication "<" normalizedInfixDirection right left)
+
+        ">=" ->
+            infer resources (Expression.OperatorApplication "<=" normalizedInfixDirection right left)
+
+        "<" ->
+            infer resources (Expression.OperatorApplication "<" normalizedInfixDirection left right)
+
+        "<=" ->
```
jfmengels: I'm thinking we could potentially convert `a >= b` to `b <= a` and `a > b` to `a < b` to normalize a bit more. What do you think?

lue-bird: isn't that exactly what is done here?

tests/Simplify/InferTest.elm
```diff
@@ -522,11 +522,28 @@ expectEqual :
...
+        actualNormal =
+            inferredNormalize actual
+    in
+    actualNormal
+        |> Expect.equal expectedNormal
+        |> Expect.onFail
+            ("expected: "
+                ++ inferredToDebugString expectedNormal
+                ++ "\nactual: "
+                ++ inferredToDebugString actualNormal
```
jfmengels: Very nice to have! :heart:

src/Simplify/Normalize.elm
```diff
@@ -448,6 +440,64 @@ addToFunctionCall resources functionCall extraArgument =
...
+                    createOperation resources "==" left right
+
+                ">" ->
+                    createOperation resources "<=" left right
+
+                ">=" ->
+                    createOperation resources "<" left right
+
+                "<=" ->
+                    createOperation resources ">" left right
```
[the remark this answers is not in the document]
lue-bird: `createOperation` does that already. I think it's more clear to only invert the operator, then let `createOperation` do the rest

Issue #398, opened 2025-12-27
Remove unnecessary arguments for directly applied lambdas (other than the first)


jfmengels:
We currently simplify

```elm
(\_ y z -> x) a b c
--> (\y z -> x) b c
```

but we don't apply the same idea to the other arguments (only the first). We should also simplify:

```elm
(\y _ z -> x) a b c
--> (\y z -> x) a c

(\y _ -> x) a <| c
--> (\y -> x) a
```

**I am looking for:**

- Someone to implement it with/for me

PR #463, opened 2026-01-20
Remove arguments passed to ignoring lambda that are not the first arg

jfmengels:
Closes https://github.com/jfmengels/elm-review-simplify/issues/398

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -17783,6 +17783,18 @@ fullyAppliedLambdaInPipelineChecks checkInfo =
                 , moduleCustomTypes = checkInfo.moduleCustomTypes
                 }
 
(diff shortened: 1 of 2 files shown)


src/Simplify.elm
```diff
@@ -21864,83 +21881,184 @@ fullyAppliedPrefixOperatorError checkInfo =
...
+                Pattern.AllPattern ->
+                    Just
+                        { pattern = pattern
+                        , argument = argument
+                        , previousPattern = previousPattern
+                        , previousArgument = previousArgument
+                        , nextPattern = List.head restOfPatterns
+                        , nextArgument = List.head restOfArguments
+                        , info =
+                            { message = "Unnecessary wildcard argument argument"
```
lue-bird: parrot. maybe "Unnecessary ignored argument" is also more clear

Issue #460, opened 2026-01-20
Remove unnecessary composition into lambda that ignores the value


jfmengels:
```elm
f >> (\_ a -> ...)
--> (\_ a -> ...)
```

jfmengels:
Apparently we already do this for `f >> always x`  but not for lambdas.

PR #464, opened 2026-01-20
Remove compositions that pipe into a lambda that ignores first argument

jfmengels:
Fixes https://github.com/jfmengels/elm-review-simplify/pull/460


lue-bird (approved):
:+1:

Issue #339, opened 2025-02-05
Transform applied composition to application


jfmengels:
I suggest we transform the following code:
```elm
-- Just reverse the composition
(f >> g) x
--> (g << f) x

-- Also if there are more compositions
(f >> g >> h) x
--> (h << g << f) x

(f >> g >> h) <| x
--> (h << g << f) x

-- Apply the current pipeline optimization that we already have
-- (the 2 last examples) even for plain function calls
(g << f) x
--> g <| f x
-- alternatively, might be nicer? Also creates less precedence problems with surrounding code
--> g (f x)

x |> (h << g << f)
--> x |> (f >> g >> h)

-- Note, only apply transformation only to the first argument
(g << f) x y
--> (g <| f x) y
-- or
--> g (f x) y

✅
x |> (f >> g)
--> x |> f |> g

✅
(g << f) <| x
--> g <| f <| x
```

And the appropriate transformations for `<<`.

I think this makes the code a bit simpler.

There's also the rationale that this would be helpful to `Simplify` itself. We've recently added the following simplification:

```elm
List.length l /= 0
--> not (List.isEmpty l)
```

In the original implementation in https://github.com/jfmengels/elm-review-simplify/pull/314, the original transformation was the following:

```elm
List.length l /= 0
--> (not << List.isEmpty) l
```

which I thought didn't look great. Changing it to not use `<<` was unfortunately more work. However, if we have the suggested transformation, then this would be an okay intermediate transformation to have. It would take an additional fix step, but it would end up the way we would recommend it.

PR #465, opened 2026-01-22
Add reversed composition checks

jfmengels:
Closes https://github.com/jfmengels/elm-review-simplify/issues/339

### DIFF
--- a/src/Simplify.elm
+++ b/src/Simplify.elm
@@ -2005,7 +2005,6 @@ All of these also apply for `Sub`.
 
 -}
 
-import Array exposing (Array)
 import Dict exposing (Dict)
 import Elm.Docs
 import Elm.Syntax.Declaration as Declaration exposing (Declaration)
@@ -18007,7 +18006,7 @@ reversedCompositionChecks checkInfo fixesFromParent node =
                 }
                 node
                 { operators = []
-                , expressions = Array.empty
+                , expressions = []
                 , parens = []
                 }
     in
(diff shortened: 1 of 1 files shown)


src/Simplify.elm
```diff
@@ -17800,116 +17854,275 @@ fullyAppliedLambdaInPipelineChecks checkInfo =
...
                         ]
                     }
-                    error.opToReplaceRange
-                    error.fixes
+                    firstOperator
+                    (expressions
+                        |> Array.toIndexedList
+                        |> List.foldl
+                            (\( index, range ) edits ->
+                                case Array.get (Array.length expressions - index - 1) expressions of
```
lue-bird: This all seems more complicated than necessary, I'd prefer
```elm
List.map2 (\selfRange oppositeRange -> Fix.replace...)
    (List.reverse expressionRangesReverse)
    expressionRangesReverse
```
and just collecting the expression ranges in a List instead of an array (ordering doesn't matter here anyway)

src/Simplify.elm
```diff
@@ -17800,116 +17854,275 @@ fullyAppliedLambdaInPipelineChecks checkInfo =
...
+                                    Just oppositeExpressionRange ->
+                                        Fix.replaceRangeBy range (checkInfo.extractSourceCode oppositeExpressionRange) :: edits
+
+                                    Nothing ->
+                                        edits
+                            )
+                            (List.concatMap (\range -> removeBoundariesFix (Node range ())) parensRanges)
```
lue-bird: constructing an empty node seems meh, maybe introduce a helper like
```elm
removeRangeBoundariesFix : Range -> List Fix
removeRangeBoundariesFix nodeRange =
    keepOnlyFix { parentRange = nodeRange, keep = rangeWithoutBoundaries nodeRange }
```

